PR #5, opened 2020-12-17
Fix literal comparison

planetarymike:
My python threw a command line error on loading this file due to a literal comparison. I believe the proper syntax is ==, not 'is'


zachariahmilby:
I think you're right. At some point PyCharm kept suggesting I change "==" to "is", and "!=" to "is not," but I've just looked it up and it does look like comparisons should use "==," whereas "is" and "is not" check to see if two things point to the same place in memory. Thanks for correcting this!

PR #6, opened 2020-12-18
Iuvs vm file sync

planetarymike:
This is code to support file sync from the VM to the local machine. Users should be able to call

`PyUVS.data.sync_data()`

in a terminal or a notebook, and have interactive prompts guide them through setting the paths to store l1b files and SPICE kernels, as well as the IUVS VM username to use in making ssh requests. These variables are stored in the newly created `user_paths.py`, which has been added to `.gitignore`

This update also adds functionality that creates a file index, which is used by find_all_l1b to speed up location of files matching a glob pattern. **Because `find_all_l1b` potentially duplicates some functionality in get_files, let's talk about what might be done to merge these routines.** Or are we OK with these existing side-by side?

More details:

Password entry is prompted once whenever sync_data is called, and the password is stored as a plain text variable for the duration of function execution. In the future setup_user_paths could be extended to allow users to store the password in plaintext in the `setup_paths.py` file, for those who are comfortable with this. 

`sync_data` works by executing a simple rsync command for the SPICE kernels, managed by pexpect.

For L1B sync, the procedure is to download a list of all available files on the VM, compare with the local file list, and rsync only the most recent version of the files. Old files are deleted following a user prompt authorizing the deletion. This differs from the way Kyle's code worked. Kyle's code made files called `excluded_files.txt` containing obsolete L1B products. **Is it ok that these `excluded_files.txt` files are no longer made?**

There are several options to control what data is synced, including a pattern option that can be used to only sync data matching a glob pattern. Setting `minorb` to a relatively recent orbit is the easiest way to speed up execution time of the function, which takes about 2 minutes if all of the VM folders are inspected.


planetarymike:
If we want a consistent naming scheme, find_all_l1b is probably not the right name. Open to suggestions!

kconnour:
I think in general I'm ok with accepting what you've written, though I haven't looked at it too thoroughly since I'm currently somewhere with slow internet. Here are some things you may consider, though if you'd rather make the utility available to sooner rather than later, I'm perfectly ok with that too.

I'm not sure how many comments / suggestions you want but here are the first things I notice that may be helpful (again, feel free to ignore them if you want): 
--- Regarding your comment, I'm perfectly fine with not having an excluded_files.txt. I just found that handy but if your code does something better (again, I haven't tested it... only looked at it) let's forget about that file.
--- Regarding organization, most of the functions in data.py are for working with data but these functions are usually for downloading data so I'd recommend separating them somehow. Although this overflows into another Github issue...
--- My view is functions should do something and therefore say what they do. I'd suggest changing functions like "latest_files" to "get_latest_files" or even "get_abs_paths_of_latest_files" so there's no confusion.
--- PEP8: line 451, 458, 605, 694, 739, 766. Import statements should be at the top of the file.
--- line 524: I believe tcsh used to be the default shell for Mac and now zsh is, so I could imagine a user may not have bash. Perhaps there's a way to just get the user's default shell to ensure you're using a valid shell? 
--- line 830: except Exception: is identical to except: as far as I know, and bare excepts are against PEP8. Perhaps you can try running your code while disconnected from the VPN? That should throw an error that you could handle and make your message more specific.
--- line 842: this is just a personal preference I suppose, but I find f strings *dramatically* easier to read. Plus they're also faster. We should strongly consider using them instead of the old style strings. 
--- General questions about sync_data: 
     - Can it sync l1a data? If so I don't immediately get that from the docstring.
     - I think by default it'll try to sync ~3TB of data which I'm not sure is desired behavior for a first time user.
     - Usually functions should do one thing and one thing only. This function does many things. Could you break up the logic into smaller functions so that this function reads almost like pseudocode?

And for setup.py
--- setup.py is a special kind of file so I don't believe it needs a docstring
--- I'd suggest changing url="https://github.com/lasp/maven_iuvs"
--- I'm ok with the authors just being the IUVS team
--- Shouldn't getpass, pexpect, and fnmatch as versioning? 
--- Why is the dependency link necessary?

planetarymike:
Eventually we'll need either Zac or Kyle to check out this branch and test it, but I'll make some changes now :)

> Regarding your comment, I'm perfectly fine with not having an excluded_files.txt. I just found that handy but if your code does something better (again, I haven't tested it... only looked at it) let's forget about that file.

I think your code downloaded everything from the VM and then added obsolete files one at a time to the excluded_files.txt so they wouldn't be synced again in later calls. This code downloads filenames only from the VM at first and then rsyncs only the newest version of the files. It's a bit slower to start syncing data but moves only the required data from the VM to the local machine.

> Regarding organization, most of the functions in data.py are for working with data but these functions are usually for downloading data so I'd recommend separating them somehow. Although this overflows into another Github issue...

No problem to separate these, but data.py is such a generic name that it doesn't describe what's happening. I propose `search.py` for file finding utilities and `download.py` for routines that interact with remote databases... what do you think? That split should be a separate PR after this one, I think.

> My view is functions should do something and therefore say what they do. I'd suggest changing functions like "latest_files" to "get_latest_files" or even "get_abs_paths_of_latest_files" so there's no confusion.

`get_latest_files` works for me. Updated.

> PEP8: line 451, 458, 605, 694, 739, 766. Import statements should be at the top of the file.

I don't understand why this is in the standard--- seems weird to pollute the global namespace and have the functions so far from their context. I'd prefer to change this once the sync routines are split out into a separate file.

> line 524: I believe tcsh used to be the default shell for Mac and now zsh is, so I could imagine a user may not have bash. Perhaps there's a way to just get the user's default shell to ensure you're using a valid shell?

Removed the bash dependency--- turns out I didn't need this. pexpect will now call whatever shell is its default based on the system.

> line 830: except Exception: is identical to except: as far as I know, and bare excepts are against PEP8. Perhaps you can try running your code while disconnected from the VPN? That should throw an error that you could handle and make your message more specific.

This was inherited from sync_data.py that Zac sent me, which has a bare except. I changed it to his because bare except catches KeyboardInterrupt and other stuff that's not an Exception. 

Switched to testing the directories earlier on and catching OSError, which is reported by pexpect when the rsync command fails.

> line 842: this is just a personal preference I suppose, but I find f strings dramatically easier to read. Plus they're also faster. We should strongly consider using them instead of the old style strings.

This is another thing inherited from the file Zac sent--- please edit if desired, I have no attachment to this line.

> General questions about sync_data:
> Can it sync l1a data? If so I don't immediately get that from the docstring.

No--- it only looks at the L1B directory.

> I think by default it'll try to sync ~3TB of data which I'm not sure is desired behavior for a first time user.

The operation can be interrupted at any time with no negative effects, but I put a warning in setup_user_paths that this could happen.

> Usually functions should do one thing and one thing only. This function does many things. Could you break up the logic into smaller functions so that this function reads almost like pseudocode?

I'm not sure what the benefit of this would be, since all of the steps need to happen every time the sync occurs and its unlikely they would need to be used in other contexts. I removed all but a small amount of redundancy via dedicated functions, splitting up the rest would involve a lot of passing of common variables through function signatures...

> And for setup.py
> setup.py is a special kind of file so I don't believe it needs a docstring

PEP 8 says it does. This problem goes away if we drop the function wrapper, as in the updated code. Not sure why there was a function wrapper there in the first place....

> I'd suggest changing url="https://github.com/lasp/maven_iuvs"

This is a thematically independent change but seems like we're doing a lot of those so OK.

> I'm ok with the authors just being the IUVS team

Since the three of us are merge managers and doing a lot of work to get this going I think it makes sense to keep our names there :)

> Shouldn't getpass, pexpect, and fnmatch as versioning?

Added a version number to pexpect. The others are python defaults so never should have appeared here, sorry!

> Why is the dependency link necessary?

Not sure, it was a holdover from the version in the repo that violated PEP8 line length reqs. I've removed it.

kconnour:
> Eventually we'll need either Zac or Kyle to check out this branch and test it, but I'll make some changes now :)

I checked it out and I must say this code looks slick. Here are my limited thoughts: 
* I pip installed PyUVS and couldn't run sync_data() until I made imports absolute. That is, I turned ".geometry" into "PyUVS.geometry" for example.
* I found the user interface really nice!
* However, I couldn't get the code to work. After inputting the requested info I got a timeout error: pexpect.exceptions.TIMEOUT: Timeout exceeded. I ran it again and got the same error so I'm not sure if I made a mistake somewhere.

> No problem to separate these, but data.py is such a generic name that it doesn't describe what's happening. I propose `search.py` for file finding utilities and `download.py` for routines that interact with remote databases... what do you think? That split should be a separate PR after this one, I think.

I like download.py and I'm fine with search.py

> > PEP8: line 451, 458, 605, 694, 739, 766. Import statements should be at the top of the file.
> 
> I don't understand why this is in the standard--- seems weird to pollute the global namespace and have the functions so far from their context. I'd prefer to change this once the sync routines are split out into a separate file.
> 
I do like PEP8 here. If I want to see if something is imported I know where to look, and I know imports aren't in places I may not expect. Plus it avoids circular imports. But anyway I'm just pointing out PEP8 here since we had previously discussed it.
 
> > line 842: this is just a personal preference I suppose, but I find f strings dramatically easier to read. Plus they're also faster. We should strongly consider using them instead of the old style strings.
> 
> This is another thing inherited from the file Zac sent--- please edit if desired, I have no attachment to this line.
> 
I shall at some point but probably not until next year. I'd rather accept this pull and then reformat things at a latter date.

> > Usually functions should do one thing and one thing only. This function does many things. Could you break up the logic into smaller functions so that this function reads almost like pseudocode?
> 
> I'm not sure what the benefit of this would be, since all of the steps need to happen every time the sync occurs and its unlikely they would need to be used in other contexts. I removed all but a small amount of redundancy via dedicated functions, splitting up the rest would involve a lot of passing of common variables through function signatures...
> 
Everything you write here alerts me that you could benefit from using a class, though it's up to you if you decide to go that route. In any case, if you could imagine anyone wanting L1a data, perhaps some of the logic could be reused in that context. But it's up to you; I'm just providing a suggestion.

planetarymike:
Do you have your system set up so that a password is not required to ssh to maven_iuvs_itf? In that case pexpect would sit around waiting for a password prompt and never receiving one...

kconnour:
Yep my system is configured in that way. Is there an easy solution?

planetarymike:
I made a change to the code to expect either a password request or some percentage transfer output so hopefully it works now?

The relative imports thing makes sense--- we'll need to make a bunch of updates to reflect this.

Totally fine with moving all the imports to the top when we move everything into a separate file.

kconnour:
I'm still having the same error running your most recent code unfortunately. Perhaps @zachariahmilby can try running it to see if it's user error or really something related to how we have our systems configured?

planetarymike:
Hmm, is it syncing files in the background at all? That would distinguish between a failure in dispatching the command vs. interacting with rsync.

Is rsync available by default in the Mac terminal? That's another point of possible failure.

Wish I had access to a Mac to sort this out...

planetarymike:
@kconnour Can you uncomment line #529 in data.py and post the rsync command(s) being issued? And paste that command into the terminal and report back on the output of the rsync command?

kconnour:
Note: Just to be sure I'm on the correct commit, I'm using the one with message "removed hard set..." Or rather, I started from there, made a change to the two relative imports and pushed that change to your fork. I've never collaboratively worked on a repo, so tell me if I did something wrong or undesirable... 

I had partial success with your new code. After running it, it began downloading files to the directory I requested! However, it timed out after a minute just like before. The result is that it downloaded files to the proper directory for about a minute before timing out and it stopped downloading additional files. When I ran it again, it picked up where it left off and downloaded files for about another minute before stopping. I really like the CLI and utility of what you've written! Just unfortunately I have something configured in a way that doesn't want to work.

kconnour:
Uninstalling and reinstalling PyUVS fixed my issues so I think I'm ready to accept this code!

Issue #7, opened 2020-12-18
Organization of routines in data.py , variables.py


planetarymike:
I'm wondering if some rearrangement of the routines is appropriate... the get_calibration_curve routine in data.py and many of the variables in variables.py refer to IUVS instrument specific quantities or procedures. Should we move these into a more descriptive submodule, like instrument.py ?

planetarymike:
Similarly, mirror_dn_to_deg in miscellaneous.py

kconnour:
I tend to think this is a good idea, although I'm little confused about your claim of IUVS specificness---it looks to me that every single function in data.py is somehow specific to IUVS. I haven't found the time to work more on my ballyhooed files classes, which may be beneficial for this sort of thing. Regardless of whether or not you like my idea, I think the more organization the better. 

I think it would be a good idea to put these functions in files that describe on what data they can act. Some functions in data.py are only valid on l1b files, whereas some seem like they can work on data from any level.

planetarymike:
I was thinking about functions specific to the instrument hardware. You're right that the whole package is specific to IUVS :)

There are definitely some other organization issues we should deal with. I was thinking about this specific issue that could result in a PR later on, so if there are others, let's open some other issues!

planetarymike:
This is closed with the incorporation of PR #10

PR #10, opened 2020-12-21
Moved data.py contents to new thematic submodules

planetarymike:
Moved file contents with minimal PEP8 modifications to new thematic subunits. Responds to #7.

Would like to get buy-in from both @zachariahmilby and @kconnour as this moves some important routines and is likely to break imports in code not tracked in the module. (All in-module imports have been updated.)


zachariahmilby (approved):
Looks fine. I don't think I have any active code that will be difficult to change to use this new organization.

kconnour:
I agree. These used to be Zac's old functions so they don't break any of my code and if they're more organized it'll be that much easier to integrate with my code in the future.

PR #12, opened 2020-12-21
Move science week stuff to time.py

kconnour:
I went ahead and moved my science week stuff to time.py since that's reasonable to me. However, I envision that at least 2/3 functions within time.py will be moved in the future, since 2/3 functions depend on spice... and the 1 function that doesn't depend on spice could potentially become part of a parent class of ScienceWeek but that's for a later time. 

I went ahead and make a tests directory for the tests I had written. I like the numpy style but I didn't make a dir previously cause I was going to wait until someone brought it up. It'd be great if tests can be run automatically but I'm not sure how to do that. 

One grumble of mine: my science week code is pretty basic and only really uses the datetime module. However, since it's now in time.py it has to import .data, which means it then has to import beta_flip from PyUVS.geometry, which means it needs to import numpy, matplotlib, cartopy, astropy, skimage, etc. to even be useable. Now ideally this won't break anything but it encumbers the lightweight code I wrote. The same is therefore true for its tests---every time I want to run tests I need to import all these modules I don't need.

I'm not exactly sure how in depth you envision we look at code. I tried to give you mostly "high level" comments on your previous request, but checking for PEP8 is by its nature digging into the weeds... anyway I'm happy to take as many detailed comments as you want to provide.


planetarymike:
Thanks, this LGTM. 

In general I think a detailed code review is not necessary for most contributions, unless the code is intended to be immediately useful to others. We should make sure stuff has decent names and is in mostly the right place. If you want a more detailed review, let me know!

I made some commits to fix some import errors and PEP8 compliance, but that's about it. My linter notes that you're missing a docstring on the class itself and on science_start_date.

With regard to where the time.py routines belong, I don't think that everything that depends on SPICE should live in spice.py. Otherwise a huge amount of geometry and visualization code would have to move there. I see spice.py as handling the basic routines to interact with SPICE kernels and load them appropriately, so that other routines can use them.

As far as the dependencies go--- this is another reason to prefer local rather than global imports ;) . But are you really worried about importing the various modules once at the beginning of the maven_iuvs module load? I can't imagine you're running in an environment that can't handle loading the modules at all, and if your class execution doesn't depend on them, having them imported shouldn't slow you down much... one way around this would be to use subdirectories to organize all the submodules, but that could get cumbersome quick.

planetarymike:
I'm gonna merge this, if you have any objections we can roll back :)

kconnour:
Sounds good! I'll make a note to myself I should add docstrings to properties. Out of curiosity, why is it better to use `datetime.date` instead of just `from datetime import date`?

planetarymike:
Either way would have worked--- in general I like to keep things encapsulated in their own namespace unless it's a real burden. I admit that typing datetime.datetime all over the place is not great though and can pollute the code. I think other parts of the code in the module needed timedelta and datetime in addition to date, so it's either import all of those routines individually or keep stuff wrapped up in the parent module.

PR #13, opened 2020-12-22
routines to sync EUVM and integrated reports data from MAVEN SDC

planetarymike:
These routines sync some ancillary data I use in making H corona quicklooks from the MAVEN SDC. Should be portable to anyone's machine, though I doubt anyone else will use these in exactly the same way. May be useful as a building block for other sync routines.

### DIFF
--- a/PyUVS/download.py
+++ b/PyUVS/download.py
@@ -1,16 +1,31 @@
-import glob
 import os
+import glob
 import subprocess
-import pexpect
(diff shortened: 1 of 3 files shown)


planetarymike:
Not sure how we should be assigning reviewers? First-come, first review?

zachariahmilby (approved):

PR #15, opened 2020-12-22
Create pull_request_template.md

planetarymike:
Adding an automatically populated pull request checklist. 

Any thoughts about what else should be added as an item?


planetarymike:
How is documentation generated? Should we add an item to regenerate the docs and push them?

zachariahmilby:
I copied the code from @kconnour, so he might want to provide more detail, but all I do is run the `make_documentation.py` file and it automatically generated everything.

zachariahmilby (approved):

PR #17, opened 2020-12-22
Update pull_request_template.md

planetarymike:
Anything else we should add here?


kconnour:
If you like my idea regarding the history then this looks like a fine set of requests to start with, and we can always add more if the need arises. I would however suggest changing `-` to `*`. Right now the file reads as one big line but `*`s would make it display properly. I only say this because I don't really know markdown that well, so if you know of a better way, by all means go for it!

planetarymike:
I typed the checklist markdown incorrectly in the original file, fixed now.

PR #19, opened 2020-12-23
Add file finding routines in class form

kconnour:
At long last here's the much ballyhooed class structure. I haven't included very many public methods but I really think this can be a skeleton from which we can add basic file handling methods. The idea is no "advanced" file routines go in this class but these classes could be input into other classes/functions for more advanced utilities. I can and plan to add several more methods, but I figured I may as well commit this code since it's useable as is. 

I included a hierarchy of classes and occasional methods slotted into the proper classes. If you don't want to work with classes directly, I added 3 helper functions that essentially just create patterns that play well with the classes. These classes are inspired by Zac's request a while ago. 

I encourage any and all input on the details of this implementation. While this commit breaks the rule of avoiding code duplication, recall we made an exception for this implementation... therefore I put the code in its own file to isolate it somewhat from the rest of the repo.

### DIFF
--- a/maven_iuvs/files.py
+++ b/maven_iuvs/files.py
@@ -1,0 +1,444 @@
+# Built-in imports
+import fnmatch as fnm
+import glob
+import os
+import warnings
+
+# 3rd-party imports
+from astropy.io import fits
+import numpy as np
+
+
+class Files:
+    """ A Files object stores the absolute paths to files, along with some file handling routines."""
+    def __init__(self, path, patterns):
+        """
+        Parameters
+        ----------
+        path: str
+            The location where to start looking for files.
+        patterns: list
+            List of strings of regex patterns to match filenames to.
+
+        Properties
+        ----------
+        absolute_file_paths: list
+            Absolute file paths of all files matching patterns in path.
+        filenames: list
+            Filenames of all files matching patterns in path.
+
+        Notes
+        -----
+        This class used glob-style matching, so a pattern of ['**/*.pdf'] will recursively search for .pdf files
+        starting from path.
+        """
+        self.__check_path_exists(path)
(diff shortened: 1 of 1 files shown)


planetarymike:
Kyle, can you reissue this PR?

It looks like once something is merged the only way to revert is to create
another pull request to revert the merge, which removes the original PR but
doesn't reopen it.

kconnour:
How do I do this? I only see a revert button, and even if I click it I don't see anything to revert.

planetarymike:
I think you need to make an entirely new pull request, with the same
content as this version, and reference this version so the history is
preserved. Super annoying, I'm not sure why it's like this...

kconnour:
I submitted a new PR with the same content. Let me know if there's something else I need to do.

planetarymike:
Hi @kconnour, looks like @zachariahmilby merged the PR without discussion again, so we need to do this over again. Sorry for the trouble!

zachariahmilby:
Sorry, I thought we were supposed to merge code we knew worked. I’ve seen Kyle’s files class in action and used it myself, so I thought it qualified.

On Dec 23, 2020, at 1:12 PM, Mike Chaffin <notifications@github.com<mailto:notifications@github.com>> wrote:



Hi @kconnour<https://github.com/kconnour>, looks like @zachariahmilby<https://github.com/zachariahmilby> merged the PR without discussion again, so we need to do this over again. Sorry for the trouble!

planetarymike:
No worries--- we're all learning how this process works. Turns out merging is irreversible and should only be done when we're absolutely sure code should be in the repo.

I think the goal of the discussion in this PR (probably #23 at this point) should be to determine what path forward we want to take with file finding routines. At the end of the day, we should end up with one set of routines, either the existing functional set or something similar, or the class-based approach in this PR. Having two sets of approved routines that do the same thing is not the best situation to end up in.

PR #21, opened 2020-12-23
Add file finding utilities in class form

kconnour:
Again, at long last here's the much ballyhooed class structure. I haven't included very many public methods but I really think this can be a skeleton from which we can add basic file handling methods. The idea is no "advanced" file routines go in this class but these classes could be input into other classes/functions for more advanced utilities. I can and plan to add several more methods, but I figured I may as well commit this code since it's useable as is.

I included a hierarchy of classes and occasional methods slotted into the proper classes. If you don't want to work with classes directly, I added 3 helper functions that essentially just create patterns that play well with the classes. These classes are inspired by Zac's request a while ago.

I encourage any and all input on the details of this implementation. While this commit breaks the rule of avoiding code duplication, recall we made an exception for this implementation... therefore I put the code in its own file to isolate it somewhat from the rest of the repo.

### DIFF
--- a/maven_iuvs/files.py
+++ b/maven_iuvs/files.py
@@ -1,0 +1,444 @@
+# Built-in imports
+import fnmatch as fnm
+import glob
+import os
+import warnings
+
+# 3rd-party imports
+from astropy.io import fits
+import numpy as np
+
+
+class Files:
+    """ A Files object stores the absolute paths to files, along with some file handling routines."""
+    def __init__(self, path, patterns):
+        """
+        Parameters
+        ----------
+        path: str
+            The location where to start looking for files.
+        patterns: list
+            List of strings of regex patterns to match filenames to.
+
+        Properties
+        ----------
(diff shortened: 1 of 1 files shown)


planetarymike:
Thanks Kyle! I'll take a look at this later today. This is a redo of PR #19.

planetarymike:
@zachariahmilby, please don't merge without consent from all reviewers. Merging eliminates the chance for changes to be made and for comments to be tracked relevant to this issue, and undoing the merge is impossible without issuing a new PR, which creates a huge mess.

We have some important decisions to make about how we're going to handle file finding moving forward, and this PR (or the next one, since we're doing this all again) is the right place to have the discussion. At the end of it we'll need to eliminate the redundant code, either what's currently in the repo or not merging this class.

planetarymike:
@kconnour, looks like you'll need to do all the work of submitting a new PR on this change again, since we need to revert this merge. Sorry for the trouble!

PR #25, opened 2020-12-24
Revised comment in call_rsync and fixed a bug in data_sync()

keimasunaga:
My first pull request :)

Comment revised because call_rsync() works not only to sync spice but also l1b data.
Do not pass call_rsync() when product data do not exist (i.e., len(prod_filenames)==0) because it fails.


maven_iuvs/download.py
```diff
@@ -109,7 +109,7 @@ def call_rsync(remote_path,
                ssh_password,
                extra_flags=""):
     """
-    Updates the SPICE kernels by rsyncing the VM folders to the local machine.
+    Updates data (e.g., L1b data and SPICE kernels) by rsyncing the VM folders to the local machine.
```
planetarymike, line 112: Thanks for noticing this!

maven_iuvs/download.py
```diff
@@ -417,11 +417,12 @@ def sync_data(spice=True, l1b=True,
 
             print('Syncing ' + str(len(files_from_production)) +
                   ' files from production...')
-            call_rsync(login+production_l1b,
-                       l1b_dir,
-                       iuvs_vm_password,
-                       extra_flags=('--files-from=' +
-                                    transfer_from_production_file.name))
+            if len(prod_filenames) > 0:
```
planetarymike, line 420: If we add this line here, we should probably add a similar line below at line 435, or (better) factor out this bit of code so that changes don't need to be made in more than one place.

 But actually my preference would be to just comment out the print statement on line 159, since running the rsync command doesn't hurt anything if no files are moved. I had this commented out originally and added it back in to do some debugging. Would that solve your issue Kei? If not, we can make the larger change to factor out the relevant code.

maven_iuvs/download.py
```diff
@@ -625,7 +626,7 @@ def sync_integrated_reports(sdc_username, sdc_password, check_old=False):
     """
 
     print("syncing Integrated Reports...")
-    
+
```
planetarymike, line 629: thanks for the PEP8 check!

planetarymike (commented):
Thanks for making this contribution, Kei! See my comments below :)

maven_iuvs/download.py
keimasunaga: Yea, I'll try to fix comments if they are not completely correct or misleading, but it's better to check my English as always.

maven_iuvs/download.py
keimasunaga: Oops it accidentally caused during debugging, but I guess it does not violate anything. I often insert import pdb; pdb.set_trace() (similar to the stop command in IDL) to check values in variables, so I could delete the spaces  when I delete the pdb command.

planetarymike: This whitespace was in my original commit, thanks for fixing it!

maven_iuvs/download.py
planetarymike: You can continue to commit to this branch and the pull request will update automatically.

You said that you get an error when the code is called with no file? What error do you get?

maven_iuvs/download.py
planetarymike: This looks good to me

maven_iuvs/download.py
keimasunaga: Ok, I'll check with this branch then. 
Below is the original error before I've added this pull request.

In [13]: maven_iuvs.download.sync_data(spice=False, l1b=True, minorb=12800, max
    ...: orb=12900)
input password for kema5665@maven-iuvs-itf: 
Fetching names of level 1B production and stage files from the VM...
                                                                               Syncing 0 files from production...
running rsync_command: rsync -trvzL --progress --files-from=/var/folders/ps/l1f7kmcx4997m2lywdkfnm7m0000gn/T/tmpf3px6esj kema5665@maven-iuvs-itf:/maven_iuvs/production/products/level1b/ /Volumes/Fenix/work/data/maven_iuvs/level1b/
---------------------------------------------------------------------------
EOF                                       Traceback (most recent call last)
~/opt/anaconda3/lib/python3.7/site-packages/pexpect/expect.py in expect_loop(self, timeout)
    110                 # Still have time left, so read more data
--> 111                 incoming = spawn.read_nonblocking(spawn.maxread, timeout)
    112                 if self.spawn.delayafterread is not None:

~/opt/anaconda3/lib/python3.7/site-packages/pexpect/pty_spawn.py in read_nonblocking(self, size, timeout)
    499         if (timeout != 0) and select(timeout):
--> 500             return super(spawn, self).read_nonblocking(size)
    501 

~/opt/anaconda3/lib/python3.7/site-packages/pexpect/spawnbase.py in read_nonblocking(self, size, timeout)
    175             self.flag_eof = True
--> 176             raise EOF('End Of File (EOF). Empty string style platform.')
    177 

EOF: End Of File (EOF). Empty string style platform.

During handling of the above exception, another exception occurred:

EOF                                       Traceback (most recent call last)
<ipython-input-13-65e421725022> in <module>()
----> 1 maven_iuvs.download.sync_data(spice=False, l1b=True, minorb=12800, maxorb=12900)

~/work/python_git/maven/lasp_iuvs/maven_iuvs/maven_iuvs/download.py in sync_data(spice, l1b, pattern, minorb, maxorb, include_cruise)
    422                        iuvs_vm_password,
    423                        extra_flags=('--files-from=' +
--> 424                                     transfer_from_production_file.name))
    425 
    426             # stage, identical to above

~/work/python_git/maven/lasp_iuvs/maven_iuvs/maven_iuvs/download.py in call_rsync(remote_path, local_path, ssh_password, extra_flags)
    163     cpl = child.compile_pattern_list(['.* password: ',
    164                                       '[0-9]+%'])
--> 165     child.expect_list(cpl)
    166 
    167     if 'password' in child.after:

~/opt/anaconda3/lib/python3.7/site-packages/pexpect/spawnbase.py in expect_list(self, pattern_list, timeout, searchwindowsize, async_, **kw)
    367             return expect_async(exp, timeout)
    368         else:
--> 369             return exp.expect_loop(timeout)
    370 
    371     def expect_exact(self, pattern_list, timeout=-1, searchwindowsize=-1,

~/opt/anaconda3/lib/python3.7/site-packages/pexpect/expect.py in expect_loop(self, timeout)
    115                     timeout = end_time - time.time()
    116         except EOF as e:
--> 117             return self.eof(e)
    118         except TIMEOUT as e:
    119             return self.timeout(e)

~/opt/anaconda3/lib/python3.7/site-packages/pexpect/expect.py in eof(self, err)
     61             if err is not None:
     62                 msg = str(err) + '\n' + msg
---> 63             raise EOF(msg)
     64 
     65     def timeout(self, err=None):

EOF: End Of File (EOF). Empty string style platform.
<pexpect.pty_spawn.spawn object at 0x1061c72d0>
command: /usr/bin/rsync
args: [b'/usr/bin/rsync', b'-trvzL', b'--progress', b'--files-from=/var/folders/ps/l1f7kmcx4997m2lywdkfnm7m0000gn/T/tmpf3px6esj', b'kema5665@maven-iuvs-itf:/maven_iuvs/production/products/level1b/', b'/Volumes/Fenix/work/data/maven_iuvs/level1b/']
buffer (last 100 chars): ''
before (last 100 chars): ' to consider\r\n\r\nsent 8 bytes  received 13 bytes  14.00 bytes/sec\r\ntotal size is 0  speedup is 0.00\r\n'
after: <class 'pexpect.exceptions.EOF'>
match: None
match_index: None
exitstatus: None
flag_eof: True
pid: 93602
child_fd: 23
closed: False
timeout: 30
delimiter: <class 'pexpect.exceptions.EOF'>
logfile: None
logfile_read: None
logfile_send: None
maxread: 2000
ignorecase: False
searchwindowsize: None
delaybeforesend: 0.05
delayafterclose: 0.1
delayafterterminate: 0.1
searcher: searcher_re:
    0: re.compile('.* password: ')
    1: re.compile('[0-9]+%')

planetarymike: oh, ok. actually this is a problem we should solve in the call_rsync function by modifying the first pexpect.expect command. I'll try to push a commit and see if it fixes your problem.

planetarymike:
Hi Kei,

I pushed some changes to the call_rsync command that should fix your issue without having to check for zero-length of production_files, let me know if this works for you.

Mike

keimasunaga:
Thanks Mike! 
I will check!

planetarymike:
I accidentally committed broken code so make sure you pull the most recent version after my last commit. Sorry!

keimasunaga:
So you made changes in this branch on the upstream repo. How can I pull this branch to my local repo in my computer? 
"git pull upstream  issue_download"  gives an error:  fatal: Couldn't find remote ref issue_download

planetarymike:
Actually the changes should be in your origin repo, so you should be able to pull from there. I was able to push to your repo because of a checkbox on the PR form "allow maintainers to make changes".

keimasunaga:
Ahh got it! Now I pulled the changes you made and checked that it solved the problem! Thanks a lot!

planetarymike:
Great! Since we're both happy I'll approve, and request approval from Kyle or Zac

planetarymike (approved):
iterated to consensus with Kei

kconnour:
It seems this code doesn't quite work on my computer from the get go. I had issues getting Python to recognize changes Mike made with a previous pull request on data download code, so I ran `pip uninstall maven_iuvs`, then checked out this branch. When I ran `python download.py` I got an error that `user_paths.py` didn't exist. I think the issue was that Python was in the processing of creating that file while it moved on to code that required that file. Every time I subsequently ran the code (when `user_paths.py` existed beforehand) it worked without issue. I also uninstalled and reinstalled this module and it worked correctly every time because `user_paths.py` existed on the the computer (or at least that's my guess of why it didn't originally work). I think it'd be easiest to make this change in this request but this issue doesn't address the original topic of the request...

keimasunaga:
Not sure if I understood correctly, but do you mean that the code works without changes made in this pull-request?
As I commented above (which is hidden in outdated tab), I got an error not due to no user_paths.py but due to something else. My original pull-request did not completely solve the error but subsequent corrections made by Mike did, at least in my local environment.

kconnour:
> Not sure if I understood correctly, but do you mean that the code works without changes made in this pull-request?

I'm not entirely sure. I still had code on my computer from when Mike and I iterated on this code during a previous pull request, and that code worked. I can't say for sure if this code did/didn't work from a clean install, but if you were having issues there was probably a bug we missed. 

> As I commented above (which is hidden in outdated tab), I got an error not due to no user_paths.py but due to something else. My original pull-request did not completely solve the error but subsequent corrections made by Mike did, at least in my local environment.

Oh ok I see that now and I'm glad the issue seems to be fixed for you. I'm curious to see if someone else has the same error I had. If no one has the same issue as me I can approve this anyway and make a note that I'll just have to run this twice after a clean install.

planetarymike:
We may want to move this whole set of features into YAML or an alternative that's more standard for storing configuration parameters, but I think that change probably belongs in another pull request. 

I think if users install as a development library the caching errors @kconnour was running into won't happen, since caching will be disabled. This would require modification of the install command to `pip install -e` rather than a vanilla pip install. That seems like an update each user needs to make, but we could update the README to suggest a dev install along with this PR.

@kconnour, if that satisfies you then can you push that commit and merge?

kconnour:
> @kconnour, if that satisfies you then can you push that commit and merge?

Sure, I'm satisfied. Since I didn't write this code I mainly wanted the authors to know about what I was seeing. I can approve this request and merge, but what should I be pushing? I haven't modified this code so I have nothing to commit.

planetarymike:
I think it can be an independent change to the README--- I'll make it now. 

I think to merge appropriately you need to submit an approving review? Not sure. Using admin powers to bypass the merge is possible but let's try to figure out the right way to do it.

kconnour (approved):
Looks good to me.

kconnour:
We have the 2 approvals, so I'd say as soon as you're happy with your minor changes you can go ahead and squash and merge. Let me know if I need to do anything else (or if I've done something wrong... this is my first approval request).

PR #26, opened 2020-12-25
Remove data_directory argument and introudce data_root_dir instead; A…

keimasunaga:
…dd level keyword

We can take advantage of user_paths.py so we do not have to input data_directory.
Also, level keyword would be useful to handle other level data (e.g., l2b)

These are my suggestion.
I thought it was a little troublesome to input a variable "data_directory" every search.
If we set data_root_dir or something (which is one layer above l1b_dir) in user_paths.py, we do not have to input the variable "data_directory" to many functions (such as get_files() and get_apoapse_files() and so on) in search module.
However, "data_directory" is used many places in this module, and these functions in this module are also used in other modules, so these changes may affect other modules.

Also, I thought the level keyword is useful (to see l2b data sometimes).

Any thoughts?

### DIFF
--- a/maven_iuvs/search.py
+++ b/maven_iuvs/search.py
@@ -8,19 +8,17 @@ from astropy.io import fits
 #  NOTE: depends on maven_iuvs.download must be encapsulated to avoid
 #  circular import
 from maven_iuvs.geometry import beta_flip
+from maven_iuvs.user_paths import root_data_dir as data_directory
 
 
-def get_files(orbit_number, data_directory,
-              segment='apoapse', channel='muv', count=False):
+def get_files(orbit_number,
+              segment='apoapse', channel='muv', level='1b', count=False):
     """Return file paths to FITS files for a given orbit number.
 
     Parameters
     ----------
     orbit_number : int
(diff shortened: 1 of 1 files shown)


planetarymike:
Thanks--- I think these commits will be superseded by changes implemented in #23, which will determine the way file search will work in the library going forward. We're going to retain all the functionality of the existing implementations and hopefully a similar syntax.

Note that imports from `maven_iuvs.user_paths` need to follow the syntax in `find_all_l1b()`, also defined in `search.py`, to avoid an `ImportError` in case the user hasn't defined `user_paths` yet.

Any issues with closing this for now, pending what happens with the other PR?

keimasunaga:
Ok, I just saw #23. Should be ok to close this PR for now, but I don't know how the class in #23 works yet. Is it possible for me to use this class?

planetarymike:
We're still working out the details, so if you wanted to use Kyle's implementation you'd have to pull from his branch, or check out the pull request. At the end of the day we'll probably be making significant changes to the existing file finding routines, but the goal is to make it as transparent as possible. Closing this for now

PR #27, opened 2021-01-04
Update README.md

planetarymike:
suggest installing in dev mode so changes to source files propagate into python without reinstallation.


kconnour (approved):

planetarymike:
Looks like we need @zachariahmilby to approve this change too before we can merge.

zachariahmilby (approved):

PR #29, opened 2021-01-23
Maven orbit image

planetarymike:
This is kind of a huge commit--- my intent was to add my maven_orbit_image routines so other could generate a 3D visualization of the orbit, but that ended up needing to pull in a bunch of time routines also to print Ls values on the images.

I don't expect anyone to read through and critique maven_orbit_image, since it's a monster. It does what I want it to do on my machine. Users can call it as is and reference it if they want to do something similar in the future. Mayavi might be tricky to install, I'm not sure... I tried to add the necessary dependencies in setup.py, but there may be non-pip dependencies. I'd be grateful if someone can check that the routine works.

Let's talk about a few of the design choices in this commit:
- moving graphics routines to a subpackage in the graphics folder--- yes, or no? I expect there will be a lot of code contributed here by lots of users so I moved the existing package into the new directory. We can do more organization later as needed.
- defining anc_dir in the packages init.py so that subroutines have access to the ancillary data they need.
- Use of an ancillary Mars Year boundaries file as a lookup table. Right now Zac uses an approximate function to place ETs into Mars Years... I could switch to doing that if someone can guide me about how to do it.

Happy to hear any suggestions!

## Pull Request Checklist


zachariahmilby (approved):
I can't pretend to understand much of how SPICE works, but these seem fine to me...

planetarymike:
The new maven_orbit_image.py is so big it's hidden by default when you look at "Files Changed", but that's where the 3D graphics and image stuff lives.

planetarymike:
Zac, I assume you're OK with the other changes in the bulleted list above?

zachariahmilby:
Yup

On Jan 26, 2021, at 4:08 PM, Mike Chaffin <notifications@github.com<mailto:notifications@github.com>> wrote:



Zac, I assume you're OK with the other changes in the bulleted list above?

planetarymike:
Great, then we.jusr need approval from Kyle and we can merge this in.

kconnour (approved):
If Zac's ok with it, I'm ok with it. In general, if a utility is primarily used by one person I'm inclined to just accept it and we can modify it as needed.

planetarymike:
Totally agreed--- I think we only need back-and-forth about modifications to existing routines, and long discussion about modifications to core routines only.

PR #30, opened 2021-01-26
tweaks to calculate_calibration_curve

planetarymike:
Hi Zac,

I made some tweaks to calculate_calibration_curve that I'd like you to review. The default behavior should be unchanged, but a new flag allows users to get one or several line calibrations if desired. 

I noticed you pull from numpy save files instead of the pipeline IDL sav file with the calibration curve in both the FUV and the MUV. This doesn't account for a 27% reduction in the FUV sensitivity we adopted in 2014, and I'm not sure where the 7 nm wavelength shift in the MUV is accounted for. I also noticed you use log interpolation of the sensitivity rather than linear. Linear is used in the IUVS IDL pipeline, what's the reason you switched to log?

Thanks!

## Pull Request Checklist
 - [n/a] New routines don't duplicate existing functionality


planetarymike:
Let me add that it doesn't matter to me if we use the original IDL save file or numpy save files for the sensitivity, as long as we get the right answer in the end. I left your code in there--- fine with me to modify that to account for the FUV issue and leave your MUV curve unchanged, since I don't use MUV data much.

zachariahmilby:
Hi Mike,

I believe I ported this code from an IDL snippet that Sonal gave me, so the use of log and no wavelength shifting were a part of that code. If I had to guess, Justin's new MUV calibration curve from 2018 probably incorporates any wavelength shift natively, so it isn't a problem anymore. As for the logarithmic interpolation, I'm not really sure. The numpy file is just a conversion of the information in the IDL file, I don't believe there are any changes to the information contained within.

I am certain that my version calculates the correct MUV calibration curve though (I have tested it directly with Sonal). Could you make a comparison plot for any periapse MUV file showing the old calibration curve made by my code and the new curve made with your changes? I am curious if they are the same.

I don't know anything about the FUV curve so I'll defer to your judgment there. I don't believe I've tested or calibrated any FUV data, and kept the sensitivity curve in the save file assuming I would get to it one day, so any changes you make to that are good!

-Zac

planetarymike:
I pushed some commits to change back to using your ancillary files, and made this plot comparing them:

Screenshot from 2021-01-26 11-29-33

I added the 1.27 multiplier for the FUV since this is needed to match what's done in the pipeline. I double checked that this matches my code, and have checked with the l1c Lya values in the past to confirm that it's needed.

For the MUV, it looks to me like the generic l1b pipeline does not use the same calibration curve as your file, so the python function will produce different cal values than the pipeline by default. Roping @planetaryjain in to comment on this--- is the newer MUV cal used somewhere else in pipeline processing?

We might consider adding a pipeline_cal flag to the function to use the l1b pipeline sensitivity curve so that users can access this if desired. Even if we don't, adding some date/origin information to the filenames would be good so that we have some idea of where they came from--- I added a comment to reflect this in the function where the files are loaded.

Mike

planetaryjain:
Hi @planetarymike,

The MUV calibration is still not used in the pipeline, but we are using this in our all MUV analysis (whether it is nightlow, aurora or airglow).  

Could you please tell me the source of the orange FUV sensitivity curve. Also, I agree with @planetarymike about naming the source file in the repo (if it is being converted to numpy save files then we should put information about original IDL sav file).

Sonal

planetarymike:
Hi @planetaryjain , I'm working on some updates to address these issues, including switching to HDF5 for the save file so that more info can be included about the origin of the various curves.

The orange FUV curve is what's in the pipeline calibration file "sensitivity update 6_9_14.sav". Blue is what happens when you multiple this by 1.27, which is done in the IDL pipeline software and is not reflected in the file.

Can you tell me when/where the orange MUV curve comes from, so I can include a note in the HDF5 file?

planetaryjain:
Hi @planetarymike,

The orange curve is the proposed updated MUV calibration IDL sav file - sensitivity update 19_10_18 prelim.sav. 

So basically, the blue is one we are currently using in our L1B pipeline?

-Sonal


> Hi @planetaryjain , I'm working on some updates to address these issues, including switching to HDF5 for the save file so that more info can be included about the origin of the various curves.
> 
> The orange FUV curve is what's in the pipeline calibration file "sensitivity update 6_9_14.sav". Blue is what happens when you multiple this by 1.27, which is done in the IDL pipeline software and is not reflected in the file.
> 
> Can you tell me when/where the orange MUV curve comes from, so I can include a note in the HDF5 file?

planetarymike:
Is Justin Deighan responsible for the creation of the file "sensitivity update 19_10_18 prelim.sav" ?

Yes, the blue is what is currently applied in the FUV pipeline.

Can you comment on the use of log vs. linear interpolation of these curves? The pipeline currently uses linear interpolation, so use of log here will result in a slight (usually invisible) discrepancy with the pipeline cal.

planetaryjain:
Yes. That is correct. Justin Deighan is responsible for the creation of the file. 

The log interpolation would probably create some issue at the very short and long wavelength range that may be the reason for the linear approach in the pipeline.   

-S

planetarymike:
@zachariahmilby , @planetaryjain Any objections to switching this to linear in the python routine, to mirror what happens in the pipeline?

planetaryjain:
I don't have any issue with that. 

-S

planetarymike:
OK, after some drama with git rebase this is ready for final review by @zachariahmilby , and then a concurrence from Kyle. 

I merged all of the calibration files into a single HDF5 with group and dataset properties that explain the origin and caveats of the calibration curves. I then removed all the old cal files from the repo--- is this OK?

Zac, this update should not affect any existing use case for calculate_calibration_curve, but you may want to double-check by pulling this branch to your machine and running some typical use cases before approving.

zachariahmilby (approved):
Looks good, Mike. No problems on my end.

planetarymike:
Cool, let's see what Kyle thinks.

kconnour (approved):
I don't really work with the calibration curve so if Zac's ok with it, I'm happy to approve it.

PR #31, opened 2021-01-27
Added geometry routines get_pixel_vec_mso and pixelcorner_avg

planetarymike:
I believe these routines do new things that aren't included in any code currently in the repo, though it looks to me like pixelcorner_avg and the existing bin_pixels_2d may have some common code that could be refactored at some point.

## Pull Request Checklist

### DIFF
--- a/maven_iuvs/geometry.py
+++ b/maven_iuvs/geometry.py
@@ -530,3 +530,170 @@ def rotation_matrix(axis, theta):
 
     # return the rotation matrix
     return matrix
+
+
+def get_pixel_vec_mso(myfits):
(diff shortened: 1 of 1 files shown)


kconnour (approved):

zachariahmilby (approved):
Looks fine to me. I've been consistently confused by the orientation between the pixel indices and the actual spatial directions.

planetarymike:
Yeah, Chris's code to generate the pixel geometry is pretty impenetrable, and there are no diagrams. When I've tried to reproduce some of the calculations I haven't been able to and I don't know what I'm doing differently...

PR #32, opened 2021-01-28
Added EUVM Lyman alpha interpolation routine.

planetarymike:
Adding a routine to compute EUVM values for some IUVS data. I added this in search because there didn't seem to be a better place, but I'm open to suggestions.

## Pull Request Checklist
 -- There are some function level imports to keep the global namespace clean, since only this routine needs them.


kconnour (approved):
I don't necessarily think these should go in the `search` module but I'll elaborate on other pull requests that I'm now just getting to.

zachariahmilby (approved):

planetarymike:
Yeah, there is probably a better place but there's some other organization of routines we need to do...

PR #34, opened 2021-02-09
Added fast pixel corner geometry calculation routines.

planetarymike:
As promised, here are some routines to do a fast calculation of the pixel corner geometry using only what's already available in the FITS file. Without a call to SPICE, these routines can have slight inaccuracies (typically much smaller than a percent), but they run on even large FITS files in a fraction of a second. Hope this helps!

## Pull Request Checklist

### DIFF
--- a/maven_iuvs/geometry.py
+++ b/maven_iuvs/geometry.py
@@ -551,8 +551,8 @@ def transform_lonlat_to_iau_vec(lon, lat):
         Array of three-vectors in IAU Mars, with a new dimension of
         size three as the last dimension of the array.
     """
-    lon = np.pi/180*np.array(lon)
-    lat = np.pi/180*np.array(lat)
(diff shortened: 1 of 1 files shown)


zachariahmilby (commented):
Hi Mike,

You might want to consider using np.radians instead of np.pi/180. They are probably the same, but since numpy has a dedication function for conversion...

planetarymike:
Thanks--- I don't think I've used those functions before, but I switched to the builtins.

planetarymike:
On my machine, these routines produce SZA and the other quantities in ~20ms for a ~500x500 file.

kconnour (approved):

planetarymike:
@zachariahmilby, any other comments or changes before you approve?

zachariahmilby (approved):

PR #35, opened 2021-02-09
Added line_fit

planetarymike:
line_fit and the other routines added here perform an LSF fit to a spectral line at a given wavelength. There are line length departures throughout line_fit and LineFitPlot, sorry. I tried to keep to 79 characters, but due to descriptive names I couldn't always do this and retain clarity in the code. Where necessary for clarity, I kept the longer lines.

## Pull Request Checklist


zachariahmilby (commented):
Hi Mike,

Question for integration.py, line 29: will these values ever change (maybe between apoapse and perhaps observations?) I think they can be read directly from the primary header, right? Might be a more robust way if there's any potential for them to vary. Otherwise most of this is stuff outside of my experience (Ly-alpha, that is...)

planetarymike:
Nice catch on my magic numbers there! Those refer to the spatial bins of the LSF that’s loaded from the file one line above. I derived the LSF from cruise data, and it only has values for those spatial bin numbers, which cover most of the airglow slit. I think elsewhere an error is thrown if any data from a FITS file on orbit falls outside those limits.

I could store these elsewhere, maybe in an HDF5 file, or make the comment a bit clearer. I should probably also archive the notebook used to create the LSF somewhere, is that the sort of thing we should include in this repo?

zachariahmilby:
Ok, sounds good. I don't think it needs to be stored anywhere, just making sure it wasn't something that could also be read directly from a FITS file header or something.

planetarymike:
Yeah, the pixel spatial geometry is read from each FITS file as needed to construct the required LSF

kconnour (approved):

planetarymike:
Anything else you'd like changed before approval Zac?

zachariahmilby (approved):

PR #37, opened 2021-02-12
File classes for review

planetarymike:
Following up from #23, this is a cleaner thread to discuss anything that needs to be changed before merging the new file classes into master. 

## Pull Request Checklist

### DIFF
--- a/maven_iuvs/__init__.py
+++ b/maven_iuvs/__init__.py
@@ -1,8 +1,3 @@
-import os as _os
-
-anc_dir = _os.path.join(_os.path.dirname(__file__),
-                        'ancillary')
-
 from . import search
 from . import download
 from . import geometry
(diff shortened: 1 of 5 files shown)


kconnour:
I'll look at this request periodically, but let me know when you have it in a finalized state and I'll give it a more thorough look.

planetarymike:
I think this is ready for review now, I don't have any pending changes to commit.

kconnour (approved):
Looks fine! @zachariahmilby, any thoughts on your end? We're just now looping you into the conversation...

zachariahmilby (approved):
I am going to go ahead and presume that you two have worked this one out...

planetarymike:
Hi Zac, you might want to confirm that renaming the file finding routines and changing the behavior to return an IUVSFITS is OK with you. Returning an IUVSFITS eliminates the need to separately import fits and open the files as a separate step. I modified the relevant search statements in your code to use the new routines, but I might have missed something.

PR #40, opened 2021-03-17
Sdc url update

planetarymike:
This is a bug fix required because the MAVEN SDC changed their directory structure without notification.

## Pull Request Checklist

### DIFF
--- a/maven_iuvs/download.py
+++ b/maven_iuvs/download.py
@@ -581,7 +581,7 @@ def sync_euvm_l2b(sdc_username, sdc_password):
     """
     print("syncing EUVM L2B...")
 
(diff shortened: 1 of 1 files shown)


kconnour (approved):

zachariahmilby (approved):
I seem to be able to access the directory with both URLs. I'll approve for the removal of the print statement, but are you sure the change of URL is necessary?

planetarymike:
It turns out this was broken because of a web configuration error at the SDC. Theoretically both should work at all times, but I pushed this before that was resolved. I'll keep the version in the PR because that one still worked even with whatever went wrong at the SDC.

PR #51, opened 2023-05-09
Added code to the initial user path setup that asks for a directory i…

emcangi:
…n which to store the full mission reprocess data.

## Pull Request Checklist


planetarymike (approved):
LGTM, eventually we should refactor this to only ask about the directories the user is trying to sync.

Issue #43, opened 2023-05-08
fix first run of sync_data


planetarymike:
Eryn ran into an issue where the IUVS VM wasn't on the list of known hosts, so the package call appeared to hang in the browser

emcangi:
A fix is included in PR https://github.com/lasp/maven_iuvs/pull/54

emcangi:
Fix included in PR https://github.com/lasp/maven_iuvs/pull/54, commit https://github.com/lasp/maven_iuvs/pull/54/commits/8acb2ecfa8243ad311f039e3f788dddbf2d7b90f

planetarymike:
Fixed by #57

PR #54, opened 2023-06-08
Echelle

emcangi:
## Pull Request Checklist
 - [X] Code Follows PEP8 guidelines (explain any exceptions)
Some lines are over length to avoid breaking in awkward positions

### DIFF
--- a/maven_iuvs/echelle.py
+++ b/maven_iuvs/echelle.py
@@ -1,0 +1,201 @@
+import os as _os
+import datetime
+import numpy as np
+import textwrap
+
+from maven_iuvs.fits_processing import get_dir_metadata, iuvs_orbno_from_fname, iuvs_filename_to_datetime, find_files_missing_geometry, find_files_with_geometry, \
+                                        iuvs_segment_from_fname
+
+def ech_isdark(fidx):
+    """
+    Identifies whether an echelle file contains dark integrations.
+
+    Parameters
+    ----------
+    fidx : dictionary
+           a single dictionary entry of metadata for some observation
+    Returns
+    ----------
+    True or False
+    """
+
+    # TODO: apply more sophisticated test to determine if really dark
+    return ('dark' in fidx['name'])
+
+def ech_islight(fidx):
+    """
+    Identifies whether an echelle file contains light (observation) integrations.
+
+    Parameters
+    ----------
+    fidx : dictionary
+           a single dictionary entry of metadata for some observation
+    Returns
+    ----------
+    True or False
+    """
+    return not ech_isdark(fidx)
+
+def find_dark_options(input_light_idx, idx_list_to_search):
+    """
+    Looks for darks which will match the observation described by input_light_idx.
+
+    Parameters
+    ----------
+    input_light_idx : dictionary
+                      a single dictionary entry of metadata for some observation
+    idx_list_to_search : list of dictionaries
+                         where each dictionary is the metadata for dark files
+
+    Returns
+    ----------
+    dark_options : list of dictionaries
+                   where each dictionary is the metadata for dark files that match input_light_idx
+    """
+    if not ech_islight(input_light_idx):
+        raise ValueError('Input file index corresponds to a dark observation, cannot find matching dark.')
+    
+    half_orbit = datetime.timedelta(hours=2)
+    dark_options = [didx for didx in idx_list_to_search 
+                    if (np.abs(didx['datetime'] - input_light_idx['datetime']) < half_orbit
+                        and didx['binning'] == input_light_idx['binning']
+                        #and didx['mcp_gain'] == input_light_idx['mcp_gain']
+                        and didx['int_time'] == input_light_idx['int_time']
+                        #and iuvs_orbno_from_fname(didx['name']) == iuvs_orbno_from_fname(input_light_idx['name'])
+                        and iuvs_segment_from_fname(didx['name']) == iuvs_segment_from_fname(input_light_idx['name'])
+                        and ech_isdark(didx))]
+    
+    return dark_options
+
+def identify_rogue_observations(idx):
+    """
+    Report on problematic observations, with either missing lights or darks, or missing data..
+
+    Parameters
+    ----------
+    idx : .npy file
+          index of observation metadata
+
+    Returns
+    ----------
+    Prints information
+    """
+    # find observations from segments where there are either lights or darks but not both
+    
+    segments = np.unique([iuvs_segment_from_fname(fidx['name'])
+                          for fidx in idx
+                          if 'orbit' in fidx['name']])
+    
+    orbits = sorted(np.unique([iuvs_orbno_from_fname(fidx['name'])
+                               for fidx in idx
+                               if 'orbit' in fidx['name']]))
+    
+    for s in segments:
+        no_issues = True
+        segment_idx = [fidx for fidx in idx
+                       if iuvs_segment_from_fname(fidx['name']) == s]
+        
+        print(f'\n{s}: ({len(segment_idx)} l1a files)')
+        for o in orbits:
+            orbit_segment_idx = [fidx for fidx in segment_idx
+                                 if iuvs_orbno_from_fname(fidx['name']) == o]
+            light_orbit_segment_flist = [fidx for fidx in orbit_segment_idx if ech_islight(fidx)]
+            dark_orbit_segment_flist = [fidx for fidx in orbit_segment_idx if ech_isdark(fidx)]
+            
+            if len(dark_orbit_segment_flist) == 0 and len(light_orbit_segment_flist) != 0:
+                print(f'  Orbit {o} light without dark')
+                no_issues = False
+            
+            if (len(dark_orbit_segment_flist) != 0 and len(light_orbit_segment_flist) == 0):
+                print(f'  Orbit {o} dark without light')
+                no_issues = False
+        
+        obs_missing_frames = [fidx for fidx in idx 
+                              if (iuvs_segment_from_fname(fidx['name']) == s
+                                  and (fidx['missing_frames'] is not None))]
+        if len(obs_missing_frames) > 0:
+            no_issues = False
+            
+            # TODO: use integrated report to check if the cutoffs are normal and due to segments ending early
+            print(f'  Frames with missing data:')
+            for fidx in obs_missing_frames:
+                if len(fidx['missing_frames']) == 1:
+                    missing_frames_string = f"{fidx['missing_frames'][0]+1}/{fidx['n_int']}"
+                else:
+                    missing_frames_string = f"{fidx['missing_frames'][0]+1}-{fidx['missing_frames'][-1]+1}/{fidx['n_int']}"
+                print(f"    {fidx['name']}: {missing_frames_string}")
+                
+        if no_issues:
+            print('  No issues.')
+
+# WEEKLY REPORT CODE ==================================================
+
+def weekly_echelle_report(weeks_before_now_to_report, root_folder):
+    """
+    Run the weekly echelle report.
+    
+    Parameters
+    ----------
+    weeks_before_now_to_report : int
+                                 number of weeks for which to run report
+    root_folder: string
+                 base folder containing all mission data in subfolders sorted by orbit
+
+    Returns
+    -------
+    None -- just updates the index files 
+    """
+    # Load the index file
+    idx = get_dir_metadata(root_folder)
+    
+    # Get data on new files 
+    # weeks_before_now_to_report = 2
+    weekly_report_datetime_start = datetime.datetime.utcnow() - datetime.timedelta(weeks=weeks_before_now_to_report)
+
+    weekly_report_idx = [fidx for fidx in idx if fidx['datetime'] >= weekly_report_datetime_start]
+    weekly_report_idx = sorted(weekly_report_idx, key=lambda i:i['datetime'])
+    weekly_report_orbit_start = iuvs_orbno_from_fname(weekly_report_idx[0]['name'])
+
+    weekly_report_idx = [fidx for fidx in idx if ('orbit' in fidx['name'] and iuvs_orbno_from_fname(fidx['name']) >= weekly_report_orbit_start)]
+    weekly_report_idx = sorted(weekly_report_idx, key=lambda i:i['datetime'])
+
+    # extend one orbit earlier to search for appropriate darks
+    weekly_report_dark_idx = [fidx for fidx in idx if ('orbit' in fidx['name'] and iuvs_orbno_from_fname(fidx['name']) >= weekly_report_orbit_start-1)]
+    weekly_report_dark_idx = sorted(weekly_report_dark_idx, key=lambda i:i['datetime'])
+    
+    # print weekly report text
+    print(f'Echelle report for {datetime.datetime.now().isoformat()[:10]}')
+    print(f'------------------------------------')
+    print(f"  covering observations after {weekly_report_idx[0]['datetime'].isoformat()[:19].replace('T',' ')} UTC")
+    print(f"                              orbit {iuvs_orbno_from_fname(weekly_report_idx[0]['name'])}+\n")
+
+    latest_orbit_with_files = iuvs_orbno_from_fname(weekly_report_idx[-1])
+    print(f'Data available through ------> orbit {latest_orbit_with_files} ({iuvs_filename_to_datetime(weekly_report_idx[-1]["name"]).isoformat()[:10]})')
+
+    geom_files = find_files_with_geometry(weekly_report_idx)
+    latest_orbit_with_geometry = iuvs_orbno_from_fname(geom_files[-1])
+    print(f'Geometry available through --> orbit {latest_orbit_with_geometry} ({iuvs_filename_to_datetime(geom_files[-1]["name"]).isoformat()[:10]})')
+
+    nogeom_files  = find_files_missing_geometry(weekly_report_idx)
+    nogeom_orbits = np.unique([iuvs_orbno_from_fname(f['name']) for f in nogeom_files if 'orbit' in f['name']])
+    print(f'  Orbits missing geometry:')
+    print('\n    '.join(textwrap.wrap(f"    {' '.join([str(orbno).rjust(5) for orbno in nogeom_orbits])}")))
+
+    weekly_lights_missing_darks = [fidx['name'] 
+                                   for fidx in weekly_report_idx 
+                                   if (ech_islight(fidx) 
+                                       and 
+                                       len(find_dark_options(fidx, weekly_report_dark_idx))<1)]
+    if len(weekly_lights_missing_darks) == 0:
+        print(f'\nAll lights have appropriate darks.')
+    elif len(weekly_lights_missing_darks) == 1:
+        print(f'\nThere is 1 light for which there is no appropriate dark:')
+        print(f'    {weekly_lights_missing_darks[0]}')
+    else:
+        print(f'\nThere are {len(weekly_lights_missing_darks)} lights for which there is no appropriate dark:')
+        for f in weekly_lights_missing_darks:
+            print(f'    {f}')
+
+    # Now list issues with each segment type
+    identify_rogue_observations(weekly_report_idx)
+    
--- a/maven_iuvs/echelle.py
+++ b/maven_iuvs/echelle.py
@@ -1,10 +1,12 @@
-import os as _os
 import datetime
 import numpy as np
 import textwrap
 
-from maven_iuvs.fits_processing import get_dir_metadata, iuvs_orbno_from_fname, iuvs_filename_to_datetime, find_files_missing_geometry, find_files_with_geometry, \
-                                        iuvs_segment_from_fname
+from maven_iuvs.fits_processing import get_dir_metadata, \
+    iuvs_orbno_from_fname, iuvs_filename_to_datetime, \
+    find_files_missing_geometry, find_files_with_geometry, \
+    iuvs_segment_from_fname
+
 
 def ech_isdark(fidx):
     """
@@ -22,9 +24,10 @@ def ech_isdark(fidx):
     # TODO: apply more sophisticated test to determine if really dark
     return ('dark' in fidx['name'])
 
+
 def ech_islight(fidx):
     """
-    Identifies whether an echelle file contains light (observation) integrations.
+    Identifies whether an echelle file has light (observation) integrations.
 
     Parameters
     ----------
@@ -36,21 +39,23 @@ def ech_islight(fidx):
     """
     return not ech_isdark(fidx)
 
+
 def find_dark_options(input_light_idx, idx_list_to_search):
     """
-    Looks for darks which will match the observation described by input_light_idx.
+    Looks for darks matching the observation described by input_light_idx.
 
     Parameters
     ----------
     input_light_idx : dictionary
-                      a single dictionary entry of metadata for some observation
+                      a dictionary entry of metadata for some observation
     idx_list_to_search : list of dictionaries
                          where each dictionary is the metadata for dark files
 
     Returns
     ----------
     dark_options : list of dictionaries
-                   where each dictionary is the metadata for dark files that match input_light_idx
+                   where each dictionary is the metadata for dark files that 
+                   match input_light_idx
     """
     if not ech_islight(input_light_idx):
         raise ValueError('Input file index corresponds to a dark observation, cannot find matching dark.')
@@ -59,17 +64,19 @@ def find_dark_options(input_light_idx, idx_list_to_search):
     dark_options = [didx for didx in idx_list_to_search 
                     if (np.abs(didx['datetime'] - input_light_idx['datetime']) < half_orbit
                         and didx['binning'] == input_light_idx['binning']
-                        #and didx['mcp_gain'] == input_light_idx['mcp_gain']
+                        # and didx['mcp_gain'] == input_light_idx['mcp_gain']
                         and didx['int_time'] == input_light_idx['int_time']
-                        #and iuvs_orbno_from_fname(didx['name']) == iuvs_orbno_from_fname(input_light_idx['name'])
+                        # and iuvs_orbno_from_fname(didx['name']) == iuvs_orbno_from_fname(input_light_idx['name'])
                         and iuvs_segment_from_fname(didx['name']) == iuvs_segment_from_fname(input_light_idx['name'])
                         and ech_isdark(didx))]
     
     return dark_options
 
+
 def identify_rogue_observations(idx):
     """
-    Report on problematic observations, with either missing lights or darks, or missing data..
+    Report on problematic observations, with either missing lights or darks, 
+    or missing data..
 
     Parameters
     ----------
@@ -80,8 +87,9 @@ def identify_rogue_observations(idx):
     ----------
     Prints information
     """
-    # find observations from segments where there are either lights or darks but not both
-    
+
+    # find observations from segments where there are either lights or darks 
+    # but not both
     segments = np.unique([iuvs_segment_from_fname(fidx['name'])
                           for fidx in idx
                           if 'orbit' in fidx['name']])
@@ -116,8 +124,9 @@ def identify_rogue_observations(idx):
         if len(obs_missing_frames) > 0:
             no_issues = False
             
-            # TODO: use integrated report to check if the cutoffs are normal and due to segments ending early
-            print(f'  Frames with missing data:')
+            # TODO: use integrated report to check if the cutoffs are normal 
+            # and due to segments ending early
+            print('  Frames with missing data:')
             for fidx in obs_missing_frames:
                 if len(fidx['missing_frames']) == 1:
                     missing_frames_string = f"{fidx['missing_frames'][0]+1}/{fidx['n_int']}"
@@ -129,6 +138,7 @@ def identify_rogue_observations(idx):
             print('  No issues.')
 
 # WEEKLY REPORT CODE ==================================================
+
 
 def weekly_echelle_report(weeks_before_now_to_report, root_folder):
     """
@@ -149,7 +159,6 @@ def weekly_echelle_report(weeks_before_now_to_report, root_folder):
     idx = get_dir_metadata(root_folder)
     
     # Get data on new files 
-    # weeks_before_now_to_report = 2
     weekly_report_datetime_start = datetime.datetime.utcnow() - datetime.timedelta(weeks=weeks_before_now_to_report)
 
     weekly_report_idx = [fidx for fidx in idx if fidx['datetime'] >= weekly_report_datetime_start]
@@ -165,7 +174,7 @@ def weekly_echelle_report(weeks_before_now_to_report, root_folder):
     
     # print weekly report text
     print(f'Echelle report for {datetime.datetime.now().isoformat()[:10]}')
-    print(f'------------------------------------')
+    print('------------------------------------')
     print(f"  covering observations after {weekly_report_idx[0]['datetime'].isoformat()[:19].replace('T',' ')} UTC")
     print(f"                              orbit {iuvs_orbno_from_fname(weekly_report_idx[0]['name'])}+\n")
 
@@ -176,9 +185,9 @@ def weekly_echelle_report(weeks_before_now_to_report, root_folder):
     latest_orbit_with_geometry = iuvs_orbno_from_fname(geom_files[-1])
     print(f'Geometry available through --> orbit {latest_orbit_with_geometry} ({iuvs_filename_to_datetime(geom_files[-1]["name"]).isoformat()[:10]})')
 
-    nogeom_files  = find_files_missing_geometry(weekly_report_idx)
+    nogeom_files = find_files_missing_geometry(weekly_report_idx)
     nogeom_orbits = np.unique([iuvs_orbno_from_fname(f['name']) for f in nogeom_files if 'orbit' in f['name']])
-    print(f'  Orbits missing geometry:')
+    print('  Orbits missing geometry:')
     print('\n    '.join(textwrap.wrap(f"    {' '.join([str(orbno).rjust(5) for orbno in nogeom_orbits])}")))
 
     weekly_lights_missing_darks = [fidx['name'] 
@@ -187,9 +196,9 @@ def weekly_echelle_report(weeks_before_now_to_report, root_folder):
                                        and 
                                        len(find_dark_options(fidx, weekly_report_dark_idx))<1)]
     if len(weekly_lights_missing_darks) == 0:
-        print(f'\nAll lights have appropriate darks.')
+        print('\nAll lights have appropriate darks.')
     elif len(weekly_lights_missing_darks) == 1:
-        print(f'\nThere is 1 light for which there is no appropriate dark:')
+        print('\nThere is 1 light for which there is no appropriate dark:')
         print(f'    {weekly_lights_missing_darks[0]}')
     else:
         print(f'\nThere are {len(weekly_lights_missing_darks)} lights for which there is no appropriate dark:')
@@ -198,4 +207,3 @@ def weekly_echelle_report(weeks_before_now_to_report, root_folder):
 
     # Now list issues with each segment type
     identify_rogue_observations(weekly_report_idx)
-    
--- a/maven_iuvs/fits_processing.py
+++ b/maven_iuvs/fits_processing.py
@@ -5,6 +5,7 @@ from pathlib import Path
 
 from maven_iuvs.search import find_files, get_latest_files
 
+
 def find_files_missing_geometry(file_index, show_total=False):
     """
     Identifies observation files with geometry
@@ -14,7 +15,7 @@ def find_files_missing_geometry(file_index, show_total=False):
     file_index : index file (.npy) 
                  dictionaries containing metadata of various observation files
     show_total: binary
-                whether to print how many files of the total the files missing geometry comprise
+                whether to print what fraction of total the missing files are
     Returns
     ----------
     no_geom: list
@@ -27,6 +28,7 @@ def find_files_missing_geometry(file_index, show_total=False):
         print(f'{len(no_geom)} of {len(all_orbit_files)} have no geometry.\n')
         
     return no_geom
+
 
 def find_files_with_geometry(file_index):
     """
@@ -47,6 +49,7 @@ def find_files_with_geometry(file_index):
 
     # print(f'{len(with_geom)} have geometry.\n')
     return with_geom
+
 
 def get_avg_pixel_count_rate(hdul, spapixrange, spepixrange, return_npix=True):
     """
@@ -94,6 +97,7 @@ def get_avg_pixel_count_rate(hdul, spapixrange, spepixrange, return_npix=True):
     
     return countrate
 
+
 def get_binning_scheme(hdul):
     """
     Gets the binning scheme for a given FITS HDU.
@@ -131,6 +135,7 @@ def get_binning_scheme(hdul):
     return {'spapix0':spapix0, 'spabinwidth':spabinwidth, 'nspa':nspa,
             'spepix0':spepix0, 'spebinwidth':spebinwidth, 'nspe':nspe,}
 
+
 def get_countrate_diagnostics(hdul):
     """
     ...description...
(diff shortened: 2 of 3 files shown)


maven_iuvs/fits_processing.py
```diff
@@ -0,0 +1,498 @@
...
+from maven_iuvs.search import find_files, get_latest_files
+
+
+def find_files_missing_geometry(file_index, show_total=False):
+    """
+    Identifies observation files with geometry
+
+    Parameters
+    ----------
+    file_index : index file (.npy)
```
planetarymike, line 15: I think the loaded index file is an array, not an .npy filename, so this might be confusing... I guess we could define a new datatype, but let's not if we can get away with it.

emcangi:
We could, I just figured it made sense to agnosticize it now. Would you prefer it that way?

planetarymike:
Right now I think that the index generation routines are agnostic to the contents of the directories, but the echelle routines that look at the index to generate the reports and file pairing don't have any protections to make sure that the files in the index are only echelle files. Until we add protections in one or both places I think the safer option is to keep all of the index generation stuff in the echelle submodule.

emcangi:
For now I just moved those functions back to the echelle file. What do you think?

planetarymike:
I think to be consistent all of the functions that are specific to echelle and the file index need to be in echelle.py . Same for functions like get_countrate_diagnostics, which is really only useful for the echelle with the pixel ranges specified.

I think the geometry functions and the average pixel countrate functions might belong in geometry.py and statistics.py / integration.py ? 

Some of the binning stuff could go into a new binning.py file? Or does another organization scheme make more sense to you?

emcangi:
Yeah, I've been having difficulty deciding how to organize since a lot of the other functions are older, the module names very general ("instrument.py"), or the functions are specific to certain data levels (e.g. l1b). 

As for the index, at this point the update_index function doesn't care what's in the folder, which doesn't seem to me to be directly tied to echelle, it's more of a folder/fits parsing function. If you insist it be part of echelle, I will put it there though. get_countrate_diagnostics should be in there though, that one I forgot about.

Overall I thought it would be reasonable to upload these new functions now and then make a more concerted effort to organize next. My instinct would be to have l1a functions and l1b functions separate, but that will probably lead to duplicated code, so the right answer is probably to sort by topics ("binning", "line fitting", "geometry" etc) and then divide those files into sections if there need to be separate functions for l1b/l1a. But I can also organize them first instead.

emcangi:
Hmmm, yeah, let's look at it next week.

planetarymike:
Any reason you prefer alphabetical order? A different order would probably be more readable, but alphabetical is probably better than what we have now.

https://stackoverflow.com/questions/10289461/what-is-a-good-way-to-order-methods-in-a-python-class

emcangi:
Mostly because I don't know what other order would be logical, and through use of the routines, we become familiar with them and remember names of ones we use, so we can find them quickly.

Open to other ordering ideas though! Could also maybe include an index comment string at the top, but I don't know offhand if that's discourged in PEP8 or 257 or whatever they call the style now.

planetarymike:
The link I shared has some decent suggestions; put the most important stuff first, followed by helper methods; group related routines by subject matter. We don’t need to be strict about this but alphabetical seems cumbersome to maintain…

emcangi:
Oh thanks, sorry I missed it due to distractions. will have a look later.

Sent from Proton Mail mobile

-------- Original Message --------

emcangi:
Fixes https://github.com/lasp/maven_iuvs/issues/43

planetarymike:
Hey Eryn,

You marked this as ready for review yesterday, but pushed some additional changes afterward. Are you done modifying these routines / should I proceed to review?

Mike

planetarymike:
Hi Eryn,

Given the room setup here at VEXAG / Venus as a System, I don’t think I’ll be able to review these changes this week: hopefully I can review and merge them next week at the EMUS meeting. Sorry I didn’t get to them last Friday!

Mike

emcangi:
I think the large ones you're looking at are commits where I organized the functions in the file according to the suggestion of top-level at the top and helper functions/lower level at the bottom. 

Not much has changed in download.py or geometry.py. In download.py, I added a few lines to correct a problem where rsync would silently fail if the local path had a space in it. In geometry I added four routines overall. In looking back at the commits, I wasn't being specific enough, thinking the changes would be evident from the commit, but I can see where this is not the case so future commit messages will be more precise. I could possibly edit the existing commit messages, although it may confuse the issue more.

planetarymike:
I think we need to split out changes that are strictly reorganization from content changes in a way that makes the diffs legible... I believe you that there are no major changes to the code files whose diffs are full of supposed changes, but it's impossible to verify this from the diffs. Any thoughts?

emcangi:
Yes, that seems like a good idea. To me the diffs are pretty clear using Sublime Merge but I bet they are annoying in the terminal. Let me look into how to remove the relevant commits...

planetarymike:
I did some additional reading on this topic and I think the problem is on my end— git diff can’t do structured diffs that would detect this kind of function moves, so reviewing changes like this in github will never work. I’ll spin up some more comprehensive git tools on my end and check if that resolves my ability to review this effectively.

planetarymike:
OK, I am now set up with a better git toolchain. The diffs are more legible, but it's still impossible to tell whether code was changed when the functions were moved--- is this the case on your end too? Clearly indicating when code was moved in a file turns out to be pretty much impossible without function level diffing, which might not exist (I couldn't find a utility to do this).

So, in the future we need to be very careful about this kind of reorganization to make merging pull requests manageable. For this PR, I think we should put unchanged code back in its original order in the files. We could consider reorganizing the code in another PR, but if it's done within the same file we'll just end up with the same problem. Thoughts?

emcangi:
I've investigated this today; the reorganization code should definitely be separate from the content changes. This was my mistake for not realizing this was a best practice and thinking I needed to implement changes as we discussed them up-thread. Additionally, I see now partially why the diffs are hard to read, even in Sublime Merge---some text that is common between functions, such as the "Parameters" and "Returns" headers in docstrings, don't get marked as additions or deletions or moves because they are the same between functions, even when I've moved functions wholesale. 

To my eye there are four seriously problematic commits (all the rest in the PR are either not polluted with both reorgs and content changes, or the pollute is really minimal):

- d2211f01 - removed some geometry-related functions from `fits_processing.py`; adds the function `get_pix_range`. 
- 70c525bc - Added the geometry-related functions removed from `fits_processing.py` to `geometry.py` and reorganized the functions, while also adding the new functions `make_tangent_lat_lon_plot`, `make_SCalt_plot`, `make_sza_plot`
- 8349f046 - Reorganized echelle.py and added some code to draw slit guidelines on the quicklooks in `echelle.py` and `line_fit_plot.py`. 
- f797b8d1 - Reorganized several files but no substantial content change: 
	- `download.py`
	- `geometry.py`
	- `integration.py`
	- `search.py`
	- `spice.py`

I think it would be difficult to do a rebase since in all of these commits, there is stuff we would like to keep and stuff we want to get rid of. (Incidentally, I don't know why d2211f01 and 70c525bc weren't the same commit since they involve movement of the same functions; I already tried to squash them in Sublime Merge, but it caused the branch to be diverged and I wasn't confident about resolving the issue so I was able to undo the squash.) 

You mentioned undoing the moves of any code that was relatively unchanged; this includes `integration.py`, `search.py`, and `spice.py`. I think the safest thing I could do to undo the reorganizations would be to `revert` these specific files within the commit f797b8d1. I believe then that the diff between f797b8d1 and the revert commit will show no changes for the files mentioned.  I'm not sure what you were using to look at the diffs, but [the --color-moved feature](https://dev.to/cloudx/how-to-color-the-moved-code-in-git-10ei) may be helpful once I revert the changes to `integration.py`, `search.py`, and `spice.py`.

What do you think of this plan? 

(Also, note I entered commit c4d827b today just in case I needed to do anything destructive on my fork of the repo, but I decided against it. It contains good and useful updates to the quicklook code though, and adding it shouldn't cause any additional complication to this issue)

emcangi:
Ok, I've now done my best to revert the problematic commits which were polluting the diffs. Have a look and let me know if we can merge this!

emcangi:
Ok, I've now done my best to revert the problematic commits which were polluting the diffs. Have a look and let me know if we can merge this!

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,1292 @@
...
+    if np.isnan(first_dark).any():
+        bad_dark_inds.append(0)
+
+    if np.isnan(second_dark).any():
+        bad_dark_inds.extend([i for i in range(1, light_data.shape[0]) if i not in bad_light_inds])
+
+    # collect all the indices of frames with some problem and remove an equivalent number of frames from the final matrix.
+    # Make sure it's a set (unique values only)
+    all_bad_inds = sorted(list(set([*nan_light_inds, *bad_light_inds, *bad_dark_inds])))
+    dark_subtracted = np.delete(dark_subtracted, all_bad_inds, axis=0)
```
planetarymike, line 877: better to return NaN for invalid frames, that way return value always has same dimensions at input light array (rule of least surprise)

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,1292 @@
...
+def pair_lights_and_darks(selected_l1a, dark_idx, verbose=False):
+    """
+    Fills a dictionary, lights_and_darks, with light and dark metadata for a given light observation file,    
+    which makes it easier to process quicklooks. Calls on find_dark_options, so any errors in dark association
+    should be fixed within that function.
+    
+    Parameters
+    ----------
+    selected_l1a : list of dictionaries
+                   selected l1a metadata entries to use
```
planetarymike, line 942: (presumably a list of light files)

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,1292 @@
...
+    Parameters
+    ----------
+    fidx : dictionary
+           a single dictionary entry of metadata for some observation
+    Returns
+    ----------
+    True or False
+    """
+
+    return 'dark' in fidx['name']
```
planetarymike, line 1049: need to add a TODO to impose a stricter test of "actually a dark"

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,1292 @@
...
+    """
+    dark_idx = [fidx for fidx in ech_l1a_idx if (('orbit' in fidx['name']) and ech_isdark(fidx))]
+    dark_idx = sorted(dark_idx, key=lambda i:i['datetime'])
+    
+    return dark_idx
+
+# Count rates ----------------------------------------------------------
+
+
+def get_avg_pixel_count_rate(hdul, spapixrange, spepixrange, return_npix=True):
```
planetarymike, line 1085: probably this should be a generic routine in a different file?

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,1292 @@
...
+    return countrate
+
+
+def get_countrate_diagnostics(hdul):
+    """
+    ...description...
+
+    Parameters
+    ----------
+    hdul : astropy FITS HDUList object
```
planetarymike, line 1138: missing/incomplete docstring

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,1292 @@
...
+    Gets Ly α countrates
+    
+    Parameters
+    ----------
+    idx_entry : string
+               folder containing observations, sorted into subfolders labeled by orbit
+
+    Returns
+    -------
+    None -- just updates the index files
```
planetarymike, line 1180: incorrect return value

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,1292 @@
...
+
+def get_dir_metadata(the_dir, new_files_limit=None):
+    """
+    Gets metadata for given set of files
+
+    Parameters
+    ----------
+    the_dir : string
+              path to directory containing observation data
+    new_files_limit :
```
planetarymike, line 1199: number of new files to add before stopping and saving the resulting index.

maven_iuvs/file_classes.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, filename):
         self.__basename = _os.path.basename(filename)
         self.__check_input_is_iuvs_data_filename()
         hdulist = HDUList.fromfile(filename, mode='readonly',
-                                   memmap=True, save_backup=False,
+                                   memmap=False, save_backup=False, # Note: Changed memmap to False because having it True precluded making quicklooks. --EMC 8/10/23
```
planetarymike, line 32: We could stop using this file_class object... it seriously slows down find_files and isn't super helpful for my workflow... immaterial to the present PR but interested in your thoughts.

maven_iuvs/fits_processing.py
```diff
@@ -0,0 +1,273 @@
...
+
+    if not (set((pixhi[:-1]+1)-pixlo[1:]) == {0}):
+        raise ValueError("Error in bin table")
+
+    pixrange = np.concatenate([[pixlo[0]], pixhi+1])
+
+    return pixrange
+
+    
+def update_index(rootpath, new_files_limit_per_run=1000):
```
planetarymike, line 242: needs to be moved to echelle unless this is going to be made generic for all FUV/MUV files

maven_iuvs/fits_processing.py
```diff
@@ -0,0 +1,273 @@
+import os as _os
+import datetime
+import numpy as np
+from pathlib import Path
+import re 
+
+from maven_iuvs.search import find_files, get_latest_files
+
+
+def get_binning_scheme(hdul):
```
planetarymike, line 10: most of the routines in this file probably belong in `binning.py', 'fits_processing' seems too generic... misfits can go into misc

maven_iuvs/geometry.py
```diff
@@ -941,3 +941,178 @@ def pixelcorner_avg(pixel_x, pixel_y, pixel_z=None,
...
+    
+    n_nan = np.sum(np.isnan(geom_quantity))
+    n_quant = np.product(np.shape(geom_quantity))
+
+    nanfrac = n_nan / n_quant
+    
+    return (nanfrac < 1.0)
+
+
+def find_files_missing_geometry(file_index, show_total=False):
```
planetarymike, line 971: echelle indexing files need to be specific to echelle

maven_iuvs/geometry.py
```diff
@@ -941,3 +941,178 @@ def pixelcorner_avg(pixel_x, pixel_y, pixel_z=None,
...
+    with_geom: list
+             metadata for files with don't have geometry
+    """
+    with_geom = [f for f in file_index if 'orbit' in f['name'] and f['geom']]
+
+    # print(f'{len(with_geom)} have geometry.\n')
+    return with_geom
+
+
+def make_sza_plot(ax, fitfile, linecolor="cornflowerblue"):
```
planetarymike, line 1016: move to graphics

maven_iuvs/graphics/line_fit_plot.py
```diff
@@ -3,36 +3,61 @@
...
+                   arange=None, prange=None, force_vmin=None, force_vmax=None):
+    """Makes a density plot of detector DN for the input FITS file and
     specified integration.
 
     Parameters
     ----------
     myfits : IUVSFITS or HDUList
-        IUVS FITS file to be plotted.
+             IUVS FITS file to be plotted.
+    image_to_plot : optional data override to use when plotting.
```
planetarymike, line 25: this probably indicates a need to split routines up: one for plotting an input array, one to extract the data from a fits file and pass is to the plotting routine. Recommend refactoring.

maven_iuvs/graphics/line_fit_plot.py
```diff
@@ -3,36 +3,61 @@
...
+              Allows for sending in externally modified arrays, 
+              e.g. with dark subtraction done.
     integration : int
-       Integration number to be plotted.
+                  Integration number to be plotted.
+    draw_slit_lines: boolean
```
planetarymike, line 30: needs to have options for FUV/MUV and echelle

maven_iuvs/graphics/line_fit_plot.py
```diff
@@ -3,36 +3,61 @@
...
+              figure and axis on which to draw.
     norm : matplotlib norm
-        Norm to be used in density plot. Overrides scale.
+           Norm to be used in density plot. Overrides scale.
     cmap : int
-        Colormap number to use from idl_colorbars package.
+           Colormap number to use from idl_colorbars package.
+    titletext : string
+                optional title to add to the image.
+    fontsizes: string
```
planetarymike, line 40: give example input?

maven_iuvs/graphics/line_fit_plot.py
```diff
@@ -50,72 +75,100 @@ def detector_image(myfits, integration=0,
 
     cmap = idl_colorbars.getcmap(cmap)
 
-    ax.set_xlim([0, 1024])
-    ax.set_ylim([0, 1024])
```
planetarymike, line 54: the original intent here was to plot the full extent of the detector and overplot the file data range... this functionality needs to be preserved.

maven_iuvs/graphics/line_fit_plot.py
```diff
@@ -3,36 +3,61 @@
...
-                   arange=None, prange=None):
-    """Makes a density plot of detector counts for the input FITS file and
+# from maven_iuvs.echelle import slit_start, slit_end
+
+
+def detector_image(myfits, image_to_plot=None, integration=0, draw_slit_lines=False, slit_lines=[346, 535],
+                   fig=None, ax=None, norm=None, cmap=109, titletext="", fontsizes="medium",
+                   scale="linear", print_scale_type=True, show_colorbar=True, show_cbar_lbl=True, labels_off=False, 
+                   arange=None, prange=None, force_vmin=None, force_vmax=None):
+    """Makes a density plot of detector DN for the input FITS file and
```
planetarymike, line 18: General concern that the scale of the changes here will break an unspecified amount of my existing code in various notebooks... in future it would be better to add new routines that preserve existing functionality, factoring out code that needs reuse, and introducing new routines for new use cases.

maven_iuvs/graphics/line_fit_plot.py
```diff
@@ -50,72 +75,100 @@ def detector_image(myfits, integration=0,
...
+    data = data / (num_frames * npixperbin)
+    
     # figure out what norm to use
     if norm is None:
-        if prange is None:
-            prange = [0, 100]
-        if arange is None:
-            arange = [np.percentile(data, prange[0]),
-                      np.percentile(data, prange[1])]
+        # Range decisions used to be here, but now it is done outside in the primary plotting routine
```
planetarymike, line 113: this is definitely a breaking change and should be handled differently...

maven_iuvs/graphics/line_fit_plot.py
```diff
@@ -50,72 +75,100 @@ def detector_image(myfits, integration=0,
...
+        # Scale decisions
+        vmin = force_vmin if force_vmin is not None else arange[0]
+        vmax = force_vmax if force_vmin is not None else arange[1]
+        
         if scale == "linear":
-            norm = mpl.colors.Normalize(vmin=arange[0],
-                                        vmax=arange[1])
+            norm = mpl.colors.Normalize(vmin=vmin, vmax=vmax)
```
planetarymike, line 119: In future, please avoid stylistic changes that don't affect the content of the code or fix PEP8 violations--- this kind of thing makes it much harder to read diffs. The rule is to keep the style of the original code unless there are substantive changes. This happens in lots of places, not just here.

maven_iuvs/integration.py
```diff
@@ -255,8 +255,17 @@ def fit_line(myfits, wavelength,
 
         for ispa in range(n_spa):
             waves = myfits['Observation'].data['WAVELENGTH'][0, ispa]
-            DN = myfits['detector_dark_subtracted'].data[iint, ispa]
-            DN_unc = myfits['Random_dn_unc'].data[iint, ispa]
+
+            try:
```
planetarymike, line 259: better to read the file identifier and select the correct treatment for l1a or l1b files...

maven_iuvs/integration.py
```diff
@@ -255,8 +255,17 @@ def fit_line(myfits, wavelength,
...
+                DN = myfits['detector_dark_subtracted'].data[iint, ispa] # l1b
+            except:
+                DN = myfits['Primary'].data[iint, ispa] # l1a
+
+            try: 
+                DN_unc = myfits['Random_dn_unc'].data[iint, ispa] # l1b
+            except:
+                DN_unc = np.zeros_like(DN) # l1a -- this is wrong obviously
```
planetarymike, line 267: better to fill with NaN

maven_iuvs/miscellaneous.py
```diff
@@ -51,3 +51,34 @@ def mirror_dn_to_deg(dn, inverse=False):
...
+def find_nearest(array, value):
+    """
+    Find the closest entry in array to value. 
+    """
+    array = np.asarray(array)
+    idx = (np.abs(array - value)).argmin()
+    return idx, array[idx]
+
+
+def get_grad_colors(L, cmap, strt=0, stp=1, mikes=False):
```
planetarymike, line 65: should be in graphics

planetarymike (commented):
Hi Eryn,

Thanks for making this much more legible--- I left some comments, let me know what you think.

Mike

maven_iuvs/echelle.py
```diff
@@ -8,6 +8,7 @@
 import matplotlib.transforms as transforms
 from astropy.io import fits
 import math 
+from tqdm.auto import tqdm
```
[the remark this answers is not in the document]
emcangi: tqdm has been added to setup.py, thanks for the catch!

maven_iuvs/constants.py
```diff
@@ -1,3 +1,5 @@
 """Physical constants used in multiple locations."""
 
 R_Mars_km = 3.3895e3  # [km]
+
+nan_color = "#424242"
```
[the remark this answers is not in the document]
emcangi: Actually, I removed it entirely because there is already a variable in `graphics.py` called `color_dict` (which I missed) that contains a gray that is close enough. Replaced instances of `nan_color` with `color_dict['darkgrey']`.

maven_iuvs/download.py
```diff
@@ -131,11 +130,13 @@ def setup_user_paths():
...
     user_paths_file.write("auto_spice_load = "+auto_spice_load+"\n")
     user_paths_file.close()
-    # now scripts can import the relevant directories from user_paths
 
+    # now scripts can import the relevant directories from user_paths
 
+ 
 def get_default_data_directory(level='l1b'):
-    """Returns default l1b_dir defined in user_paths.py, creating the
+    """
```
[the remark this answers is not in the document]
emcangi: Yes, the original code (perhaps written by you or Kyle or both) contained the summary lines starting on the same line as the opening triple quote (mostly, I think). In some of my functions I started the summary on the next line because that looks better to me without really thinking about it/checking what PEP actually says. In undoing some of the messy commits, I tried to move these back to their original location on the same line as """ to ensure line numbers would match so that the overall diff would be cleaner for you [and others] to read. 

I did just check PEP and it says that either is acceptable:

> The summary line may be on the same line as the opening quotes or on the next line.

I'm currently using SublimeLinter with flake8 conventions. We can discuss to agree on a convention and then I can make a new commit later that just focuses on cleaning up any formatting issues in the docstrings/other comments!

maven_iuvs/download.py
```diff
@@ -482,11 +483,6 @@ def sync_data(spice=True, level='l1b',
         else:
             raise ValueError('local_dir must be l1a or l1b')
 
-    # Check to be sure that the vm is in the known_hosts file.
```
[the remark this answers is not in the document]
emcangi: I've looked into it and I have no idea. I think this must be a mistake--I've added it back in since it should definitely be there.

maven_iuvs/download.py
```diff
@@ -558,6 +555,9 @@ def sync_data(spice=True, level='l1b',
             # save the files to rsync to temporary files
             # this way rsync can use the files_from flag
             transfer_from_production_file = tempfile.NamedTemporaryFile()
+            # save files to sync to a local file that can be accessed for troubleshooting
```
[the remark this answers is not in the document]
emcangi: Oh, these comment lines were used when I was still diagnosing #56. I've removed these lines.

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,1292 @@
...
+                    missing_frames_string = f"{fidx['missing_frames'][0]+1}/{fidx['n_int']}"
+                else:
+                    missing_frames_string = f"{fidx['missing_frames'][0]+1}-{fidx['missing_frames'][-1]+1}/{fidx['n_int']}"
+                print(f"    {fidx['name']}: {missing_frames_string}")
+                
+        if no_issues:
+            print('  No issues.')
+
+
+# QUICKLOOK CODE ======================================================
```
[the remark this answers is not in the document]
emcangi: Created graphics/echelle_graphics.py, can rename to echelle.py if desired but thought that it might be confused with the other echelle.

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,1292 @@
...
+    show_DN_histogram : boolean
+                        Can be turned on to show a histogram of all the pixel counts.
+                        Useful for determining where to set prange, but since prange is ideally passed in,
+                        this means you will want to run this iteratively/manually.
+    verbose : boolean
+              whether to print feedback messages
+              
+    Returns
+    ----------
+    Completed quicklook figure.
```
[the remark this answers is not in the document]
emcangi: Updated the docstring and cleaned up the return values so that they are all strings rather than some strings and some booleans.

maven_iuvs/echelle.py
emcangi: Clarified this docstring

maven_iuvs/echelle.py
emcangi: Added

maven_iuvs/echelle.py
emcangi: I didn't write this one or put it in here, but happy to move it. How about moving it to geometry?

maven_iuvs/echelle.py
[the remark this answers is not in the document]
emcangi: Fixed by ecdef50a6424c0982fa48d2380b0896f03bd4539

maven_iuvs/constants.py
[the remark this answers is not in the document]
emcangi: Fixed by 2e510edc07dda7e866884b8f11f1cf76892d0571

maven_iuvs/download.py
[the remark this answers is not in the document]
emcangi: Fixed by bf5ed3a31fdd88b6eb6eb728dc473e3f0bda45ef

maven_iuvs/download.py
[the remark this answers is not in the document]
emcangi: Fixed by bf5ed3a31fdd88b6eb6eb728dc473e3f0bda45ef

maven_iuvs/echelle.py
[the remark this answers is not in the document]
emcangi: Fixed by ae415e4a60f788a69c1d123945a1e143fd23e89a

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,1292 @@
...
+    date : datetime object, or list of datetime objects
+           If a single datetime object of type datetime.datetime() or datetime.date() is entered, observations matching exactly are returned.
+           If a list is entered, observations between the two date/times are returned. A -1 may be passed in the second position to indicate to run to the end.
+           Whenever the time is not included, the code will liberally assume to start at midnight on the first day of the range 
+           and end at 23:59:59 on the last day of the range.
+    segment: an orbit segment to look for. "outlimb", "inlimb", "indisk", "outdisk", "corona", "relay" etc
+
+    Returns
+    ----------
+    selected_lights
```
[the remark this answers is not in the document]
emcangi: Fixed by d81b483549da109ff45b25607960cdaf97a7345b

maven_iuvs/echelle.py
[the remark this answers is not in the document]
emcangi: Fixed by d81b483549da109ff45b25607960cdaf97a7345b and ae415e4a60f788a69c1d123945a1e143fd23e89a

maven_iuvs/echelle.py
emcangi: Fixed by ae415e4a60f788a69c1d123945a1e143fd23e89a

maven_iuvs/echelle.py
emcangi: Fixed by d81b483549da109ff45b25607960cdaf97a7345b

maven_iuvs/echelle.py
emcangi: Fixed by d81b483549da109ff45b25607960cdaf97a7345b

maven_iuvs/echelle.py
emcangi: Fixed by d81b483549da109ff45b25607960cdaf97a7345b

maven_iuvs/echelle.py
emcangi: Fixed by d81b483549da109ff45b25607960cdaf97a7345b

maven_iuvs/echelle.py
emcangi: Fixed by d81b483549da109ff45b25607960cdaf97a7345b

maven_iuvs/file_classes.py
emcangi: I've used it in a few places, but I have no huge attachment. It is nice that it has some methods to obtain certain ephemera/data easily but it's also not enormously important. I'd like to table it for now or at least introduce a new commit to remove it, as long as that wouldn't bother anyone...

maven_iuvs/fits_processing.py
emcangi: Fixed by 6467166340944ca6a69510b1a6725e519a1a501e

maven_iuvs/fits_processing.py
emcangi: Fixed by 6467166340944ca6a69510b1a6725e519a1a501e

maven_iuvs/geometry.py
emcangi: Fixed by c361af4d4d7026605a880d550706d37d7c4378c2

maven_iuvs/geometry.py
emcangi: Fixed by c361af4d4d7026605a880d550706d37d7c4378c2

maven_iuvs/graphics/line_fit_plot.py
emcangi: Fixed by 19ff4ea22866d14108ef61c8ecf5b25a8d7f81d7

maven_iuvs/graphics/line_fit_plot.py
emcangi: Can we table this for a future commit/PR? I haven't interacted with those datasets for a while so I don't have them downloaded right now for experimenting.

maven_iuvs/graphics/line_fit_plot.py
emcangi: Fixed by 713fe2542d534540d67550f08173bd1e21f6abfc

maven_iuvs/graphics/line_fit_plot.py
emcangi: Fixed by 19ff4ea22866d14108ef61c8ecf5b25a8d7f81d7

maven_iuvs/graphics/line_fit_plot.py
emcangi: Fixed by 19ff4ea22866d14108ef61c8ecf5b25a8d7f81d7

maven_iuvs/graphics/line_fit_plot.py
emcangi: Fixed by 19ff4ea22866d14108ef61c8ecf5b25a8d7f81d7

maven_iuvs/integration.py
emcangi: Fixed by 1669a94a7aa155787effe1d66b032aecf75395d3

maven_iuvs/integration.py
emcangi: Fixed by 1669a94a7aa155787effe1d66b032aecf75395d3

maven_iuvs/miscellaneous.py
emcangi: Fixed by 6467166340944ca6a69510b1a6725e519a1a501e

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,886 @@
...
+    nan_dark_inds : array
+                    Indices of dark frames in which there are nans.
+    total_frames : int
+                   total number of good frames which were used to create the 
+                   co-added image. Returned so the value can be added to 
+                   quicklooks.
+    """
+
+    # Retrieve dark frames for subtraction
+    first_dark, second_dark = get_dark_frames(dark_fits)
```
planetarymike, line 270: I think assuming that there is always a first and second dark will cause problems when processing the very earliest mission data, but we can deal with that later.

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,886 @@
...
+
+    # check darks are valid
+    if np.isnan(first_dark).any() & np.isnan(second_dark).any():
+        raise Exception(f"Missing critical observation data: no valid darks")
+
+    # dark subtraction
+    dark_subtracted, nan_light_inds, bad_light_inds, nan_dark_inds = subtract_darks(light_fits, first_dark, second_dark)
+    
+    # check that we still have valid data
+    if dark_subtracted.shape[0] == 0: # Some files may have no acceptable frames
```
planetarymike, line 280: exceptions on returned data should be raised inside the function that returns the data

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,886 @@
...
+
+    # Retrieve dark frames for subtraction
+    first_dark, second_dark = get_dark_frames(dark_fits)
+
+    # check darks are valid
+    if np.isnan(first_dark).any() & np.isnan(second_dark).any():
+        raise Exception(f"Missing critical observation data: no valid darks")
+
+    # dark subtraction
+    dark_subtracted, nan_light_inds, bad_light_inds, nan_dark_inds = subtract_darks(light_fits, first_dark, second_dark)
```
planetarymike, line 277: my preference would be to call get_dark_frames() inside subtract_darks() unless this function (coadd_lights) needs  to know about the dark values themselves

maven_iuvs/download.py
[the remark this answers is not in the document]
planetarymike: Let's open an issue on this

maven_iuvs/echelle.py
planetarymike: it probably belongs in 'integration' if it belongs anywhere; 'geometry' is specifically routines for handling spacecraft and pointing location information.

maven_iuvs/echelle.py
```diff
@@ -0,0 +1,886 @@
...
+           containing the count rates and number of pixels included 
+           in areas on the detector covering:
+           (1) H Ly alpha emission;
+           (2) H Ly alpha background;
+           (3) D Ly alpha emission;
+           (4) D Ly alpha background.
+
+
+    """
+    Hlya_spapixrange = np.array([ech_slit_start, ech_slit_end])
```
planetarymike, line 663: need to add an exception to abort if file is not ech, since the pix ranges make no sense for FUV or MUV data

maven_iuvs/file_classes.py
planetarymike: Happy to change all the loading routines to use fits.open() instead; IUVSFITS adds a lot of overhead processing time.

maven_iuvs/graphics/graphics.py
```diff
@@ -1304,3 +1327,102 @@ def reset_symlog_labels(fig, axes):
         # set tick padding above the label
         for tick in ax.get_xaxis().get_major_ticks():
             tick.set_pad(11)
+
+
+def make_sza_plot(ax, fitfile, linecolor="cornflowerblue"):
```
planetarymike, line 1332: would be nice if ax was an optional argument so these could be called directly on a fits file, but that's not needed now.

maven_iuvs/graphics/line_fit_plot.py
```diff
@@ -50,72 +146,56 @@ def detector_image(myfits, integration=0,
...
         if arange is None:
-            arange = [np.percentile(data, prange[0]),
-                      np.percentile(data, prange[1])]
+            arange = [np.percentile(data_to_plot, prange[0]),
+                      np.percentile(data_to_plot, prange[1])]
+
+        # Allows overriding the vmin and vmax; for instance, if you 
+        # have multiple externally-created plots that you want to
+        # display with a shared colorbar.
+        vmin = force_vmin if force_vmin is not None else arange[0]
```
planetarymike, line 164: seems like arange and force_vmin,max are synonymous?

maven_iuvs/instrument.py
```diff
@@ -47,6 +47,9 @@
 slit. This is out of 1024 pixels (index 1023) for a 1024x1024 pixel
 detector."""
 
+ech_slit_start = 346  # Starting pixel of slit in echelle mode
```
planetarymike, line 50: slit position for H/D Lyman alpha

planetarymike (commented):
Looks great--- I added some small comments, let me know what you think. Hoping code changes will be significantly less onerous in the future

maven_iuvs/echelle.py
emcangi: Right now it throws an error if there's only 1, returns the 2 if there are 2, and returns 2 with the second being an average of 2:end if there's > 2. I can change the unpacking to a tuple at a later point though

maven_iuvs/echelle.py
emcangi: is it not enough to already be in the echelle module? Wouldn't we thus want to apply similar error handling in other functions?

maven_iuvs/instrument.py
emcangi: Oh yeah, thanks. You would think I would know this having looked at the images 293752 times.

maven_iuvs/echelle.py
emcangi: Fixed by 8e993bb2abcc2337088f44ee0ad8e4006fedcc1f

maven_iuvs/echelle.py
emcangi: Fixed by 1fd9de2318dba503f971986065f883b8a930c043

maven_iuvs/graphics/graphics.py
emcangi: Easy enough, fixed by 48e969aa8240f4d5c25b2e7a76aeeb796a9e5b50

maven_iuvs/graphics/line_fit_plot.py
emcangi: Yes, must be an artifact, fixed by 07835d00fd301d96b8d508a211c52cee7710a049

maven_iuvs/instrument.py
emcangi: Fixed by 0170b4e961f6c74d1ac2a1ea7d407fde769d4baf

maven_iuvs/download.py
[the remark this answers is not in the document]
emcangi: Issue https://github.com/lasp/maven_iuvs/issues/58 is opened, resolving this conversation

maven_iuvs/echelle.py
emcangi: Started on this with bcddfdf576c7739e451c3c8ae3a396c4dd0193ae

maven_iuvs/file_classes.py
emcangi: Done in ced635c7bcaa29b50074b1e9d4643fb3929406f1, there weren't that many in use, mostly I had used it in echelle.py

maven_iuvs/echelle.py
planetarymike: Yeah, probably fine to just trust the user in this case.

planetarymike:
still working on this? happy to review now or later

emcangi:
I think I've addressed every comment, but it's hard to tell because of the way it displays. Take another final look?

planetarymike:
LGTM, thanks for persevering through these changes!

planetarymike (approved):
LGTM

PR #55, opened 2023-08-03
added H quicklook code from maven_iuvs_chaffin

planetarymike:
This is an initial commit of some very messy (but working!) code that generates H Lyman alpha quicklooks. Much to improve before this can be merged, I think.

## Pull Request Checklist

### DIFF
--- a/maven_iuvs/graphics/H_quicklook/__init__.py
+++ b/maven_iuvs/graphics/H_quicklook/__init__.py
@@ -1,0 +1,3 @@
+# list of scripts to import
+
+from .get_orbfiles_and_times import get_orbfiles_and_times
--- a/maven_iuvs/graphics/H_quicklook/apoapse.py
+++ b/maven_iuvs/graphics/H_quicklook/apoapse.py
@@ -1,0 +1,273 @@
+from .plot_defaults import *
+import numpy as np
+from maven_iuvs.geometry import get_pixel_vec_mso, pixelcorner_avg
+
+def quicklook_apoapse(fig=None,orbno=None, observations=None,
+                      orbit_ax=None, orbit_coords=None, map_ax=None, to_iau_mat=None,
+                      colormap=None, cmapnorm=None, apoapse_axes=None,
+                      plot_brightness=None):
+    apoapse_obs = [obs for obs in observations if obs['obsid']=='apoapse' and obs['n_int']>1 and obs['segment']=='apoapse' and not obs['echelle'] and not obs['filename']=='']
+    #there's nothing to do for an apoapse observation without a file except put it on the orbit, which we already did.
+    
+    #if there are no apoapse observations, turn off all the axis frames and return
+    if len(apoapse_obs)==0:
+        apoapse_axes[0].set_axis_off
+        apoapse_axes[2].set_axis_off
+        return apoapse_axes
+    
+    #otherwise, continue drawing the observations  
+
+    #get the MSO pixel vecs for each observation
+    all_pixel_vecs_mso=[]
+    for idx, obs in enumerate(apoapse_obs):
+        pixel_vecs_mso=get_pixel_vec_mso(obs['fits'])
+        all_pixel_vecs_mso.append(pixel_vecs_mso)
+        
+    #
+    # plot 1: image of the planet and swaths
+    #
+
+    #planet was drawn when axes were set up
+    
+    #draw bounding box of each swath on top of planet
+    all_pixel_corners_2d=[]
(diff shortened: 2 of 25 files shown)


planetarymike:
note that some care is needed to get this working on your local system--- Mike with consult with Eryn on path forward.

planetarymike:
Here is my giant list of to-dos from the notebook where I originally developed this code:

----------------------
to-do

Continue tweaks to fit_line:
   * improve flatfield
   * improve MUV subtraction of apoapsis --- sonal knows how to do this

Plotting tweaks:
 * deal with 24 periapse obs cases like on orbit 1694
   * higher numbers of periapse files should be from nadir-looking aurora campaigns: 
         (per Justin: "A check for this type of nadir viewing auroral scan is that it should have 12 spatial bins and 3 integrations per file. There can be regular limb scans in other parts of the periapse segment.")
                                                                                                 ^^^^early versions of this sequence have 7 spatial bins
   *  there are indeed some files with the usual number of ints/bins --- need to display only these on the viz
 * Ancillary plot stuff
   * add a colorbar for MCS dust
   * add more labels to ancillary plots
   * add tau=1 and line center brightness (Ly a) to EUVM plot (probably requires interpolation)
 * width of swaths on apoapse orbits > 11000: 
   * MRH point is on near side of planet even though observation extends past limb. Draw these at location of limb instead.
 * Check printed altitudes on limb plots are using the right y-coordinate (may be backward on inbound segment)
 * Units on EUVM
 * horizontal alignment of periapsis swaths: make sure this is aligned with along-orbit direction. Right now it's a fixed based on spatial binning
    
File tweaks:
 * remove old/bad files from stage that are on the wrong orbit:
   * wtf is going on with orbits like 8240? Looks like every observation is doubled or tripled
     * this happens for orbits 8235-8250
   * due to a CenturyLink outage in early 2019. Bad files are stage tagged and can be removed
 * write some checker routine to be run in sync_latest_files
   * did this and I have some stage files that were never made by production and are also not on the PDS
   * sent a list of these to Sonal (September 2020 sometime?)

Bugs/corner cases:
 * orbit 855: serious issues with apoapse images, likely due to line integration routine
 * orbit 333: HV turns off before observation ends, even though file reports it on.
   * What to do? implement total counts per image filter? Average counts < 0?
 * file mvn_iuv_l1b_inspace-orbit00341-ech_20141202T034550_v13_r01.fits.gz seems to be assigned the wrong orbit, orbit 341 already has muv and fuv obs. Orbit 340 has an echelle obs already?
   * This appears to happen many times in the orbit range 0-1000
 * some apoapse files use nonlinear binning. Ex: orbit 386
 * issues with parsing integrated report file on orbit 972 and others
   * Possibly due to safe mode entry?
 * orbit 110 apoapsis: "ValueError: x and y arguments to pcolormesh cannot have non-finite values or be of type numpy.ma.core.MaskedArray with masked values"
   * this is because of NaN values in PIXEL_VEC --- why are these present???
   * make a list of orbits where these occur, see if there are any recent examples
 * NaNs in periapse / other geometry
   * Same issue as above for apoapsis, likely there is a bug in the mirror kernel
   * sometimes the whole map is gray? Gonna guess this is related to the above
 * v03 files are present intermittently in the data--- do these contain the same data as later version files?
 * check date alignment of ancillary values--- EUVM? these look to not be centered on the orbit
 * MAYBE instead of not making a file, print a blank canvas with the orbit and a warning that no files were found 
   * that way if there's no file there was definitely an error

Wrap up and package integrated report/filename comparator

Code to select only orbits where certain subsets are populated (apoapse, side segment, etc.)

---------------------
shortly after that

Timeline plots in new right panel:
 * depends on ly-a brightness being saved somewhere
 * timelines of periapsis/side segment brightness
 * something similar for apoapsis?

---------------------
longer term

 * print special event text if stellar occ, deep dip, conjunction, caveats file, etc.
    ** need time ranges for these events --- could get from Ops calendar
    ** there are also several lists in the caveats file

Handle corner cases
 * convert generic raise to specific python error
 * handle bad pointing for apoapsis
   * like in orbit 3498
   * not clear that this can be fixed

Lower priority tweaks:
 * map stuff
   * add label for peri+side segment scans
   * decide how to show+label side segment scans
   * add better+fast limb detection for apoapsis swath boundaries
   * label apoapsis boundaries
 * store filenames in metadata only and use the extra space for something else?

More recent ancillary datasets:
 * MCS especially--- any way to get this from the PDS?
 * do SWIA/EUVM processing locally?
    
Add generic hook for an input function to generate plot intensities for each observation
 * allows others to use the code to make plots of their own

Issue #60, opened 2024-02-16
Echelle quicklooks attempt to save figure and logfile when no savefolder is specified


emcangi:
If `savefolder` is none, the quicklook code still tries to create a path name for the figure file (`ql_filepath` variable). It also will go ahead making a logfile even if savefolder is none, and the log file will go in some random folder on the user's drive.

PR #61, opened 2024-02-16
Fix issue where the quicklook code will attempt to create a pathname …

emcangi:
…for the quicklook figure and write a logfile even if savefolder is none. Now, code will properly assign figure and log to savefolder if it exists; if it does not, no save path will be defined, no log file will be written, and the figure will just be shown.

Fixes https://github.com/lasp/maven_iuvs/issues/60. 

## Pull Request Checklist


planetarymike (approved):
LGTM

PR #63, opened 2024-03-15
Bug fixes related to dark subtraction, co addition, and orbit numbers:

emcangi:
- Add a mode to allow searching for the orbit number in the full file path. This requires regular expressions.
- Clarify dark subtraction code so that it's obvious if the 0th frame is good or not (previously, the 0th frame was excluded from i_all and therefore also i_good, even if it was a valid frame with a valid dark).
- Explicitly return the number of good frames for use elsewhere, e.g. in coaddition.
- Explicitly return "light_frames_with_nan_dark", whereas previously these were included in "nan_dark_inds" which is not clear.
- Allow coadd_lights to not return the list of indices for bad frames if desired.
- add ['name'] index to weekly_report_idx[n] in weekly report code (where n is some index of the file metadata index to explicitly pass iuvs_orbno_from_fname just the filename, rather than the entire index entry (which is then transformed to a string).

## Pull Request Checklist
 - [x] New routines don't duplicate existing functionality (no new routines)

### DIFF
--- a/maven_iuvs/echelle.py
+++ b/maven_iuvs/echelle.py
@@ -54,15 +54,15 @@ def weekly_echelle_report(weeks_before_now_to_report, root_folder):
     print(f"  covering observations after {weekly_report_idx[0]['datetime'].isoformat()[:19].replace('T',' ')} UTC")
     print(f"                              orbit {iuvs_orbno_from_fname(weekly_report_idx[0]['name'])}+\n")
 
-    latest_orbit_with_files = iuvs_orbno_from_fname(weekly_report_idx[-1])
-    print(f'Data available through ------> orbit {latest_orbit_with_files} ({iuvs_filename_to_datetime(weekly_report_idx[-1]["name"]).isoformat()[:10]})')
+    latest_orbit_with_files = iuvs_orbno_from_fname(weekly_report_idx[-1]['name'])
+    print(f"Data available through ------> orbit {latest_orbit_with_files} ({iuvs_filename_to_datetime(weekly_report_idx[-1]['name']).isoformat()[:10]})")
 
     geom_files = find_files_with_geometry(weekly_report_idx)
     try:
         latest_orbit_with_geometry = iuvs_orbno_from_fname(geom_files[-1])
-        print(f'Geometry available through --> orbit {latest_orbit_with_geometry} ({iuvs_filename_to_datetime(geom_files[-1]["name"]).isoformat()[:10]})')
+        print(f"Geometry available through --> orbit {latest_orbit_with_geometry} ({iuvs_filename_to_datetime(geom_files[-1]['name']).isoformat()[:10]})")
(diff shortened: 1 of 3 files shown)


maven_iuvs/miscellaneous.py
```diff
@@ -122,21 +131,28 @@ def orbit_folder(orbit):
...
 
     Returns
     -------
     orb_string : int
                  orbit number
     """
-    orb_string = str(fname).split('orbit')[1][:5]
-    return int(orb_string)
+    if search_full_path:
+        return int(re.search(orbno_RE, fname).group(0))
```
planetarymike, line 152: seems to me that using os.path.basename would be clearer and platform-independent…

maven_iuvs/echelle.py
```diff
@@ -54,15 +54,15 @@ def weekly_echelle_report(weeks_before_now_to_report, root_folder):
     print(f"  covering observations after {weekly_report_idx[0]['datetime'].isoformat()[:19].replace('T',' ')} UTC")
     print(f"                              orbit {iuvs_orbno_from_fname(weekly_report_idx[0]['name'])}+\n")
 
-    latest_orbit_with_files = iuvs_orbno_from_fname(weekly_report_idx[-1])
-    print(f'Data available through ------> orbit {latest_orbit_with_files} ({iuvs_filename_to_datetime(weekly_report_idx[-1]["name"]).isoformat()[:10]})')
+    latest_orbit_with_files = iuvs_orbno_from_fname(weekly_report_idx[-1]['name'])
```
planetarymike, line 57: where is this “name” element defined in your committed code? Does this code run on your machine or are there uncommitted files?

planetarymike (commented):
Hi Eryn,

Found some time to look at this while waiting for my flight.

-Mike

maven_iuvs/echelle.py
emcangi: "name" is a key in the echelle file index, it's not a separate variable or anything

maven_iuvs/miscellaneous.py
emcangi: sounds good! Made the change

maven_iuvs/echelle.py
planetarymike: my bad, I forgot this was in the index file!!

maven_iuvs/miscellaneous.py
planetarymike: committed?

emcangi: Sorry, thought I pushed but hadn't. Fixed by [6488aa7](https://github.com/lasp/maven_iuvs/pull/63/commits/6488aa7a09b1698c3a6508eee4bfd5b68241f8c6)!

planetarymike (approved):
LGTM

PR #64, opened 2024-04-29
Slight improvements to echelle quicklooks

emcangi:
- Change wspace parameter so that geometry plot y-labels don't overlap the nearby axes
- Change default fontsize for make_one_quicklook to be larger for easier legibility
- Use the median frame as a stopgap to avoid looking at cosmic rays until the cosmic ray subtraction routine is incorporated
- Simplify the way font sizes are set based on the qualitative description
- Replace some hard-coded font sizes with the dynamically adjusting sizes
- Remove some artifact comments that are no longer relevant; add one comment to explain the use of d_geo
- Fix error where the index at which to plot the Lyman alpha guideline would be erroneously calculated as np.argmax(spectrum), leading to some intense cosmic rays pretending to be Lyman alpha. Instead, calculate said index as np.argmax(spectrum) within a narrow wavelength window centered on the recorded wavelength bin most similar to Lyman alpha (doing it this way allows for the case where the greatest emission might not be exactly at the location of the closest wavelength to Lyman alpha)

## Pull Request Checklist

### DIFF
--- a/maven_iuvs/echelle.py
+++ b/maven_iuvs/echelle.py
@@ -278,6 +278,16 @@ def coadd_lights(light_fits, dark_fits):
     return coadded_lights / total_frames, [nan_light_inds, bad_light_inds, nan_dark_inds], total_frames
 
 
+def median_light_frame(light_fits, dark_fits):
+    """
+    Obtains a median light frame, which can be useful for displaying the observations without cosmic rays, etc.
+    """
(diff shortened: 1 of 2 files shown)


emcangi:
Would like to incorporate these changes into the main repo so I can merge them into the base for the l1a to l1c pipeline branch as well--some modifications in that branch will enable changes to the quicklooks, so I'd like to wait to PR the l1a-->l1c pipeline code until this quicklook update is incorporated

PR #67, opened 2025-01-15
Fix errors in data cleanup and expand capabilities of several functions

emcangi:
## Pull Request Checklist

It's been a little too long since the last PR, will aim for future PRs to be more discrete. In the meantime, I'd like to add this stuff now; the following PR will be quite soon after this one and include a refactor of some of the fitting pipeline code so that it's more modular and doesn't have to be called just in the function that generates l1c data products.

An executive summary of some major changes in this set:

1. Adds some geospatial information to the file metadata indices, such as min/max local time, SZA, and latitude/longitude
2. Update the data cleanup routines to fix a major error that was generating fake emissions and deleting real ones due to an indexing mistake
3. Add extra plotting capabilities and improve existing line fitting plots
4. Add a reprocessing.py module which consolidates ad hoc code I've used to run the full mission reprocesses and validate results
5. Add MRH altitude information to quicklooks and make the lat/lon plots easier to read
6. Add a function that generates a CSV of light and dark pairs, which consolidates ad-hoc code I've been using to create keys like this for PDS deliveries
7. Add IPH fitting support, which is not currently working likely due to improper use of the LSF to fit the shape; can be improved later.

### DIFF
--- a/maven_iuvs/echelle.py
+++ b/maven_iuvs/echelle.py
@@ -1281,9 +1281,15 @@ def convert_l1a_to_l1c(light_fits, dark_fits, light_l1a_path, savepath, calibrat
                                                     fitter=fitpackage, approach=approach, livepts=livepts, hush_warning=hush_warning) 
 
         # Create a convenient dictionary which can be used with a plotting routine
-        fit_params_for_printing = {'area': round(fit_params[0]), 'area_D': round(fit_params[1]), 
-                                    'lambdac': round(fit_params[2], 3), 'lambdac_D': round(fit_params[2]-D_offset, 3), 
-                                    'M': round(fit_params[3]), 'B': round(fit_params[4])}
+
+        if np.isnan(fit_params).all():
+            fit_params_for_printing = {'area': fit_params[0], 'area_D': fit_params[1], 
+                                        'lambdac': fit_params[2], 'lambdac_D': fit_params[2]-D_offset, 
+                                        'M': fit_params[3], 'B': fit_params[4]}
+        else:
+            fit_params_for_printing = {'area': round(fit_params[0]), 'area_D': round(fit_params[1]), 
+                                        'lambdac': round(fit_params[2], 3), 'lambdac_D': round(fit_params[2]-D_offset, 3), 
+                                        'M': round(fit_params[3]), 'B': round(fit_params[4])}
         
         # Construct a background array from the fit which can then be converted like the spectrum
         bg_fit = background(wavelengths, fit_params_for_printing['M'], fit_params_for_printing['lambdac'], fit_params_for_printing['B'])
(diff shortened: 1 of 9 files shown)


maven_iuvs/echelle.py
```diff
@@ -201,26 +201,34 @@ def downselect_data(light_index, orbit=None, date=None, segment=None, lat=None,
```
planetarymike, line 201: did you want to add comments for the new arguments?

maven_iuvs/echelle.py
```diff
@@ -810,6 +957,14 @@ def get_dir_metadata(the_dir, geospatial=False, new_files_limit=None):
     remove_from_idx = np.setdiff1d(idx_fnames, most_recent_fnames)
     new_idx = [i for i in idx if i['name'] not in remove_from_idx]
 
+    # NEW: UPDATE WITH MISSING INFO - don't run on new index, just on existing entries
```
planetarymike, line 960: nice, glad this is added now

planetarymike (commented):
@emcangi Are you aware of any changes that would affect code outside of the echelle package? I haven't done a line-by-line review, but would like to look at any such code in detail if there is any.

emcangi:
> @emcangi Are you aware of any changes that would affect code outside of the echelle package? I haven't done a line-by-line review, but would like to look at any such code in detail if there is any.

I tried to keep this in mind and not make any breaking changes, but will try to check. Since I haven't used any other packages in a while, it's not as straight forward for me to test, but I can investigate.

planetarymike:
No need to do detailed checks— I don’t think many common functions have been changed, so the most this could affect is someone’s notebook (mine) using, say, `line_fit_plot` . If you’re not aware of any other potential issues I’m ready to approve.

maven_iuvs/echelle.py
emcangi: Yes, now here: https://github.com/lasp/maven_iuvs/pull/67/commits/f006aea93e7b09ffb4882f8f7ba716aa8ec80b5b

emcangi:
line_fit_plot has had an optional argument updated. See: 06be8347c85435d972da62c63dfb6a03b57bdbe0 

However, the change should be silent. If you don't pass in `cmap` in your code, it will still default to idl_colorbars #109. The change as written now allows a user to pass in a named colormap in addition to an idl_colorbars integer.

Should still work the same!

I've paged through other commits and don't see anything referencing non-echelle code.

emcangi:
@planetarymike , please check the following two commits: 
https://github.com/emcangi/maven_iuvs/commit/f19787667f28d0d505d42ad63fae74cb9f0e3dd4 (older one)
and
https://github.com/lasp/maven_iuvs/pull/67/commits/0b996b91dfc405241ca5e0005d68d69233db96f6

and let me know if you agree with the logic!

planetarymike:
Your argument in 0b996b9 looks good, but you mentioned that number twice--- was there another number I should look at?

emcangi:
I had a typo and you responded too soon, check the post again

planetarymike:
I think either approach is OK as long as it's clear to the reader of the quicklooks what's going on. In the second case with NaNs, the question is whether the second dark is really so different from the first dark that it's better to show the raw unsubtracted data rather than use the first dark... my guess is using the first dark is better than doing nothing, but it's been so long since I've looked at this problem that I can't recall.

emcangi:
Sounds good, I think it is now clearer. Previously it was silently pretending that a second dark did exist when it did not. 

I've added one more commit above that now uses the 1st dark if the 2nd doesn't exist, but will still note on the quicklook if the dark is bad. There's not a lot of files with only 1 dark frame anyway and they're all before orbit 1000, so this affects a small amount of the data.

planetarymike:
OK, that makes sense--- I'll go ahead and approve!

planetarymike (approved):

PR #70, opened 2025-03-25
Add IPH fitting, enable SSH key authentication with VM, encapsulate l1a-->l1c code, and make other minor improvements.

emcangi:
## Pull Request Checklist
 - [X] Code Follows PEP8 guidelines (explain any exceptions)
 - [X] New routines don't duplicate existing functionality
 - [X] Original authors of modified routines have been added as reviewers

Main changes: 

1. Encapsulate convert_l1a_to_l1c() into smaller subroutines, which makes it easier to call them separately for testing and data exploration
2. Add SSH key authentication capability to sync_data() for connecting to the VM (user must manually add their public key after ssh'ing in, though)
3. Add IPH fitting routines (still not perfect--difficult to get the parameters right)
4. Created a function to convert Mars date to UTC time, which was something I needed once at some point, so I stole the calculation from the MCD by inspecting the page with dev tools :D

Minor changes:

1. Fix various bugs (see commit messages)
5. Improved residual axes bounds
6. Replaced some deprecated code (deprecated in the Python modules, not ours) -- utcnow() and np.product()


maven_iuvs/echelle.py
```diff
@@ -1415,297 +1456,342 @@ def convert_l1a_to_l1c(light_fits, dark_fits, light_l1a_path, savepath, calibrat
...
+            IPH_fit_array[i, :] = IPH_fit
+        else:
+            fit_params, I_fit, fit_1sigma, *_ = result_vec
+
+        I_fit_array[i, :] = I_fit
+        fit_unc_list.append(fit_1sigma)
+        fit_params_list.append(fit_params)
+
+    # Error control
+    if np.isnan(fit_params).all():
```
planetarymike, line 1584: a more strict criterion would be to fail if any of the integration fits return nan--- did you consider adopting this condition?

maven_iuvs/echelle.py
```diff
@@ -1415,297 +1456,342 @@ def convert_l1a_to_l1c(light_fits, dark_fits, light_l1a_path, savepath, calibrat
...
-            spec_ph_s_bg_sub = spec_ph_s - bg_ph_s
-        else:
-            spec_ph_s_bg_sub[:] = np.nan # TODO: Update this if needed to handle problem files.
+    return bg_fits
+
+
+def compute_ph_per_s_data(light_fits, spectrum, fit_params, bg_fits):
+    """
+    The echelle l1c files typically report a quantity referred to in IDL pipeline as bright_data_ph_per_s, 
+    e.g. photons per second. This function computes this value so that it can continue to be reported.
```
planetarymike, line 1651: photons per second or photoevents per second?

maven_iuvs/echelle.py
```diff
@@ -2086,10 +2174,11 @@ def lineshape_model(params, wavelength_data, binedges, theCLSF, BU_bg, fit_IPH=F
...
-        total_brightness_IPH = params[-2]
-        central_wavelength_IPH = params[-1]
-        interpolated_CLSF_IPH = interpolate_CLSF(central_wavelength_IPH, binedges, theCLSF)
-        normalized_line_shape_IPH = np.diff(interpolated_CLSF_IPH) # Unitless
+        total_brightness_IPH = params[-3] 
+        central_wavelength_IPH = params[-2]  # TODO: Can calculate based on spacecraft ephemeris and pass in for the starting guess/requirement.
+        width_IPH = params[-1]
+        CDF_IPH = sp.stats.norm.cdf( (binedges - central_wavelength_IPH) / width_IPH ) #  TODO: Should also be able to calculate this, which is the width of the Gaussian.
+        normalized_line_shape_IPH = np.diff(CDF_IPH)
         IPH_fit = total_brightness_IPH * normalized_line_shape_IPH
```
planetarymike, line 2182: This seems like it should work to me--- maybe you need some additional heuristics to guess good initial values based on the observed spectrum?

maven_iuvs/echelle.py
```diff
@@ -2378,9 +2481,9 @@ def line_fit_initial_guess(wavelengths, spectrum, H_a=20, H_b=170, D_a=80, D_b=1
 
     # Total flux of H and D initial guess: get by integrating around the line. Note that the H bounds as defined
```
planetarymike, line 2482: maybe note that this total is in DN, not kR?

planetarymike (commented):
Hi Eryn,

A few notes embedded in your code--- the IPH component looks correct at a glance, so I guess you need to implement some additional heuristics to get good initial guesses for the parameters, or rely on pymc / dynesty for the fit.

maven_iuvs/time.py
```diff
@@ -56,6 +56,38 @@ def utc_to_sol(utc):
...
+    sol : float
+          The decimal sol date. If -1, sol will be reset to the maximum
+          number of sols.
+    
+    Returns
+    -------
+    UTC datetime object.
+    """
+    # constants
+    ns = 668.6  # number of sols in a Mars year
```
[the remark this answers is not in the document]
emcangi: Hahaha, yeah. I inspected the code on the MCD using the developer console and stole it.

maven_iuvs/echelle.py
```diff
@@ -1282,130 +1266,187 @@ def get_ech_slit_indices(light_fits):
...
+                  kwargs which may be passed to the plotting routines to control plot appearance.
+    
+    Returns
+    ----------
+    Null - writes out an l1c file if the option is turned on.
+    """
+
+    # Set number of integration frames to fit 
+    # ===============================================================================================
+    ints_to_fit = {"first": 1, "all": get_n_int(light_fits)}.get(ints_to_fit, get_n_int(light_fits))
```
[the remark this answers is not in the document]
emcangi: Yeah. Fixed with https://github.com/lasp/maven_iuvs/pull/70/commits/4173cf1abe3f99cff50e1903ba2ed333af04b82d

maven_iuvs/echelle.py
emcangi: I think this was written this way because it used to be that the routines would return nan for the D fit sometimes (but still get a valid H fit) but now it properly returns a float. Now a single nan is probably appropriate to fail the fit

emcangi: Fixed by https://github.com/lasp/maven_iuvs/pull/70/commits/6cf9de7a0afc25ea0394948c9912069b02dd32a7

maven_iuvs/echelle.py
emcangi: The variable names communicate this but sure, it can be in the comment. Fixed by 27e272c0b844ce4141a300613e0396440848fac8

maven_iuvs/echelle.py
emcangi: Yeah probably b83055b4db3f81b640a650d3dd1c788fb8859d32

maven_iuvs/time.py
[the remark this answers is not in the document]
emcangi: 5965ad4f68f4c482c845e735e26337dff811f50b

maven_iuvs/echelle.py
emcangi: To respond to above comment, not even Dynesty was able to fit these for some reason

emcangi:
@planetarymike I think I'm happy with this for now. the IPH fitting is working, but can be improved with the updates we've talked about (prediction of parameters, etc). My goal is to get these changes here incorporated and then incorporate the updates in much smaller PR's, and hopefully create a separate branch to keep track of graphics updates.

maven_iuvs/echelle.py
```diff
@@ -1931,26 +2136,29 @@ def prior_transform(u, param_bounds):
...
+                      [pig[6]-1e4, pig[6]+1e4], # background slope
+                      [pig[7]-1e4, pig[7]+1e4] # Background offset
                      ] 
                    ]
 
+        print(f"Initial guess for the IPH will be: area {ptf_args[0][2]}, " 
+                + f"center {ptf_args[0][4]}, width {ptf_args[0][5]}")
+
         if approach=="dynamic":
             dsampler = d.DynamicNestedSampler(loglikelihood, prior_transform, len(ptf_args[0]),
```
planetarymike, line 2163: does the sign argument to loglikelihood need to change for dynesty?

maven_iuvs/echelle.py
```diff
@@ -1931,26 +2136,29 @@ def prior_transform(u, param_bounds):
...
         # List of arguments for loglikelihood
         loglike_args = [wavelengths, edges, CLSF, spec, unc, 1, BU_bg]
 
         # List of arguments for prior_transform
-        ptf_args = [ [[param_initial_guess[0]/1000, param_initial_guess[0]*1000], # Total DN, H
-                      [param_initial_guess[1]/1000, param_initial_guess[1]*1000],  # Total DN, D
-                      [param_initial_guess[2]-0.5, param_initial_guess[2]+0.5], # H lyman alpha center in nm
-                      [param_initial_guess[3]-1e4, param_initial_guess[3]+1e4], # background slope
-                      [param_initial_guess[4]-1e4, param_initial_guess[4]+1e4] # Background offset
+        ptf_args = [ [[pig[0]/10, pig[0]*10], # Total DN, H
```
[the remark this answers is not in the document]
emcangi: it is already within the if statement as far as I can tell. Here's what it looks like on my end (sorry, line numbers have changed slightly due to other uncommited changes)

Screenshot From 2025-07-09 13-02-57

Screenshot From 2025-07-09 13-04-50

maven_iuvs/echelle.py
emcangi: Fixed in 4629870a59d41724507a2ff402c35f9a9f594b41

maven_iuvs/graphics/echelle_graphics.py
```diff
@@ -34,7 +35,7 @@
 
 # QUICKLOOK CODE =========================================================================================
```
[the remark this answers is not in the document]
emcangi: They seem to be working, and can be fixed anyway later if bugs are found

maven_iuvs/echelle.py
```diff
@@ -2054,55 +2278,52 @@ def lineshape_model(params, wavelength_data, binedges, theCLSF, BU_bg, fit_IPH=F
...
-        total_brightness_IPH = params[-2]
-        central_wavelength_IPH = params[-1]
-        interpolated_CLSF_IPH = interpolate_CLSF(central_wavelength_IPH, binedges, theCLSF)
-        normalized_line_shape_IPH = np.diff(interpolated_CLSF_IPH) # Unitless
-        IPH_fit = total_brightness_IPH * normalized_line_shape_IPH
-        fitsum += IPH_fit
-    else:
-        IPH_fit = None
+    # IPH model uses a basic Gaussian for now
+    CDF_IPH = sp.stats.norm.cdf( (binedges - central_wavelength_IPH) / width_IPH )
```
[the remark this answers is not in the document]
emcangi: This will be tracked in https://github.com/lasp/maven_iuvs/issues/78

emcangi:
@planetarymike Have a look at the comments above, and resolve as you approve

maven_iuvs/echelle.py
```diff
@@ -1714,39 +1909,40 @@ def convert_l1a_to_l1c(light_fits, dark_fits, light_l1a_path, savepath, calibrat
...
+    # Now call IDL
+    if open_idl is True:
         os.chdir(idl_pipeline_folder)
-        commands = f'''
-                    .com write_l1c_file_from_python.pro
-                    write_l1c_file_from_python, '{light_l1a_path}', '{savepath}', '{brightness_csv_path}', '{ph_per_s_csv_path}'
-                    ''' 
-
         proc = subprocess.Popen("idl", stdin=subprocess.PIPE, stdout=subprocess.PIPE, text="true")
+        proc.stdin.write(".com write_l1c_file_from_python.pro")
```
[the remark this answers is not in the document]
planetarymike: I think leaving it out for now is OK since this is a python repo but maybe in the future the maven-iuvs-itf and this repo have one another as submodules (is that even possible?? probably)

maven_iuvs/echelle.py
planetarymike: I think if the default sign has been changed to minus, the function should be called negloglikelihood

emcangi: There's no default, it uses -1 for dynesty and +1 for scipy. Since there is no obvious negative value in it as written I think calling it negloglikelihood would be confusing

planetarymike: something like keeping the current implementation of loglikelihood() and implementing negloglikelihood() as

> `def negloglikelihood(*args, **kwargs):`
> `     return -loglikelihood(*args, **kwargs)`

emcangi: Ah, this is different from the previous message and I like this solution better.

maven_iuvs/echelle.py
```diff
@@ -2261,15 +2264,14 @@ def loglikelihood(params, wavelength_data, binedges, CLSF, data, uncertainty, s,
         A single value which represents either the log-likelihood (if negative)
         or the fit "badness" if positive.
     """
-    s = s / abs(s)  # Enforce s to be -1 or 1 since python doesn't have sign.
 
     # Now do the model 
     DN_fit, *_ = lineshape_model(params, wavelength_data, binedges, CLSF, BU_bg) 
 
     # Fit the model to the existing data assuming Gaussian distributed photo events 
     L = np.sum((data - DN_fit)**2 / (2*(uncertainty**2) ) )
```
planetarymike, line 2272: I hate to prolong this discussion, but I think loglikelihood needs to have a minus sign in it here, which gets cancelled out by negloglikelihood. (Will also need to update which function is called by scipy minimize vs dynesty)

emcangi: You forced me to review my obs and stats notes but you're right, the LL for a Gaussian has a minus sign. This makes it frustratingly unintuitive but we are stuck with it.

planetarymike: The way it makes intuitive sense to me is that log likelihood represents the log of a probability, and probabilities are always smaller than one, so the log is negative.

Approving!

emcangi: hadn't yet pushed updates when you wrote most recent comments. All are now fixed in 3bad7013aae1dda5b1104b64b052c651dbb69643. Please review and check that it's all sorted!

planetarymike (approved):
LGTM!

Issue #71, opened 2025-06-02
New Install MAVEN IUVS does not set up user_paths.py


JadenStrom:
Fresh installation generates the following error

> ---------------------------------------------------------------------------
> ModuleNotFoundError                       Traceback (most recent call last)
> Cell In[4], line 2
>       1 get_ipython().run_line_magic('cd', '[/Users/jadenstrommer/Desktop/MAVENIUVS/maven_iuvs](http://localhost:8888/lab/tree/Desktop/MAVENIUVS/Desktop/MAVENIUVS/maven_iuvs)')
> ----> 2 import maven_iuvs as iuvs
> 
> File [~/Desktop/MAVENIUVS/maven_iuvs/maven_iuvs/__init__.py:7](http://localhost:8888/lab/tree/Desktop/MAVENIUVS/~/Desktop/MAVENIUVS/maven_iuvs/maven_iuvs/__init__.py#line=6)
>       2 # keep this above the relative imports below, some depend on this
>       3 # definition
>       4 anc_dir = _os.path.join(_os.path.dirname(__file__),
>       5                         'ancillary')
> ----> 7 from . import search
>       8 from . import download
>       9 from . import geometry
> 
> File [~/Desktop/MAVENIUVS/maven_iuvs/maven_iuvs/search.py:13](http://localhost:8888/lab/tree/Desktop/MAVENIUVS/~/Desktop/MAVENIUVS/maven_iuvs/maven_iuvs/search.py#line=12)
>       9 from astropy.io import fits
>      11 #  NOTE: depends on maven_iuvs.download must be encapsulated to avoid
>      12 #  circular import
> ---> 13 from maven_iuvs.geometry import beta_flip
>      16 def find_files(data_directory=None,
>      17                recursive=True,
>      18                use_index=None,
>      19                count=False,
>      20                **filename_kwargs):
>      21     """Return fits files for a given glob pattern.
>      22 
>      23     Parameters
>    (...)     58 
>      59     """
> 
> File [~/Desktop/MAVENIUVS/maven_iuvs/maven_iuvs/geometry.py:8](http://localhost:8888/lab/tree/Desktop/MAVENIUVS/~/Desktop/MAVENIUVS/maven_iuvs/maven_iuvs/geometry.py#line=7)
>       6 from skimage.transform import resize
>       7 import pkg_resources
> ----> 8 from maven_iuvs.instrument import slit_width_deg
>      11 def beta_flip(hdul):
>      12     """
[...]

PR #74, opened 2025-06-02
Fix import errors when user_paths.py does not exist

JadenStrom:
## Pull Request Checklist
 
This addresses issue #71

### DIFF
--- a/maven_iuvs/download.py
+++ b/maven_iuvs/download.py
@@ -220,7 +220,7 @@ def call_rsync(remote_path,
 
     # Add some code to handle a case where spaces in the local folder path will cause rsync to fail silently
     if " " in local_path:
-        local_path = local_path.replace(" ", "\ ")
+        local_path = local_path.replace(" ", "\\ ")
 
     rsync_command = " ".join(['rsync -trvzL',
                               progress_flag,


planetarymike:
Yeah, you'll need to create separate branches for each PR. One change, one branch.

planetarymike:
You should be able to modify your IDE to automatically warn you about any PEP8 violations, otherwise don't worry about it too much.

JadenStrom:
Alright, sounds good. Thank you.
________________________________

JadenStrom:
Okay, I can look into the PyCharm features and see if there is any mention of those words. If not, I can just disregard it for the time being if that's alright? Or should I try to find some other workarounds to find them?
________________________________

emcangi:
Pycharm should have Pylint available: https://plugins.jetbrains.com/plugin/11084-pylint 

I forgot to clarify that these are tools for "linting" which is the process of making sure that things are tidy and correct (https://en.wikipedia.org/wiki/Lint_%28software%29)

its easiest to get started with linting right away since it keeps code readable. you will get warnings about line lengths but those I think are fine to ignore for now since the whole codebase has lots of lines that are too long. Anything else though is good to keep track of.

PR #79, opened 2025-07-16
Iph improvements

### LINKED ISSUE #78: Use better IPH fitting parameters
- [ ] IPH offset should be the fit parameter relative to the rest wavelength of Lyman alpha
- [ ] Add prediction of IPH wavelength 
- [ ] Add prediction of IPH width

emcangi:
## Pull Request Checklist
 
Initial PR addressing Issue #78.

FYI only the most recent commit is actually important. Unclear why the earlier 4 are even on this branch.


emcangi:
As of this morning, remaining task is to get the IPH brightness onto the quicklooks.

emcangi:
IPH brightness now added to QL in 49e215541c43a47085199ea2bb3b2bc9ad5a36d4. Ready for a final review

maven_iuvs/echelle.py
```diff
@@ -1607,34 +1634,41 @@ def fit_flat_data(light_fits, spectrum, data_unc, bad_frames=None,
...
 
         if BU_bg is not None:
             BU_bg_i = BU_bg[i, :]
         else:
             BU_bg_i = None
 
-        result_vec = fit_H_and_D(initial_guess, wavelengths, spectrum[i, :], light_fits, theCLSF, unc=data_unc[i, :], BU_bg=BU_bg_i, **kwargs)
-            
+        # Determine whether line-of-sight minimum ray height is large enough to fit an IPH component
+        fit_IPH_component = check_whether_IPH_fittable(light_fits)
```
planetarymike: technically this is duplicating work because the full FITS file is checked on every integration loop, but eh

maven_iuvs/echelle.py
```diff
@@ -1542,17 +1546,40 @@ def flatten(light_fits, processed_data):
...
-                  calibration="new", return_each_line_fit=False, ints_to_fit=1,  
+_fit_parameter_names = ['total_brightness_H',  # DN
+                        'total_brightness_D',  # DN
+                        'total_brightness_IPH',  # DN
+                        'central_wavelength_H',  # nm
+                        'central_wavelength_IPH',  # nm  
+                        'width_IPH',  # nm
+                        'background_m',
+                        'background_b']
+_unc_parameter_names = ['unc_total_brightness_H',
```
planetarymike: should probably use a list comprehension here for safety: `['unc_'+pname for pname in _fit_parameter_names]`

maven_iuvs/echelle.py
```diff
@@ -1892,10 +1908,10 @@ def writeout_l1c(light_l1a_path, dark_l1a_path, l1c_savepath, light_fits, binnin
...
     yMRH = math.floor((yMRH-this_dict['ycH'].values[0])/this_dict['NbinsY'].values[0]) #  to get an integer value like IDL does.
 
-    H_brightnesses = [fit_params_list[i]["area_H"] for i in range(n_int)]
-    D_brightnesses = [fit_params_list[i]["area_D"] for i in range(n_int)]
-    H_1sig = [fit_unc_list[i]["unc_H"] for i in range(n_int)]
-    D_1sig = [fit_unc_list[i]["unc_D"] for i in range(n_int)]
+    H_brightnesses = [fit_params_list[i]['total_brightness_H'] for i in range(n_int)]
+    D_brightnesses = [fit_params_list[i]['total_brightness_D'] for i in range(n_int)]
+    H_1sig = [fit_unc_list[i]['unc_total_brightness_H'] for i in range(n_int)]
+    D_1sig = [fit_unc_list[i]['unc_total_brightness_D'] for i in range(n_int)]
```
planetarymike, line 1964: we should probably report the fitted IPH parameters in the l1c file, right?

maven_iuvs/echelle.py
```diff
@@ -2070,24 +2120,34 @@ def fit_H_and_D(pig, wavelengths, spec, light_fits, CLSF, unc=1, IPH_bounds=(Non
...
-                                       bounds=[(None, None), (None, None), 
-                                               (0, None), 
-                                               (121.55, 121.58), IPH_bounds, 
-                                               (0.0001, 0.02), 
-                                               (None, None), (None, None)]
+                                       bounds=[(None, None), # DN_H
+                                               (None, None),  # DN_D
+                                               (None, None), # DN_IPH
+                                               (121.55, 121.58),  # λ H
+                                               (pig[4]-IPH_wv_spread/2,
```
planetarymike: I think +/- 5 km/s may be safer, this is +/- 2.5 km/s, right?

maven_iuvs/echelle.py
```diff
@@ -2298,55 +2365,54 @@ def lineshape_model(params, wavelength_data, binedges, theCLSF, BU_bg):
...
-    total_brightness_D = params[1] 
-    total_brightness_IPH = params[2] 
-    central_wavelength_H = params[3] # nm
-    central_wavelength_D = params[3] - D_offset # nm
-    central_wavelength_IPH = params[4]  # TODO: Can calculate based on 
-                                         # spacecraft ephemeris and pass in for 
-                                         # the starting guess/requirement.
-    width_IPH = params[5]  # TODO: Should also be able to calculate this, 
-                            # which is the width of the Gaussian. 
+    param_dict = make_fit_param_dict(params, BU_bg=BU_bg)
```
planetarymike: nice, this is more elegant. Changing this function to a class with static class members for the names / bounds / etc. is probably the move to avoid having things at the global scope, but the current implementation is good enough.

maven_iuvs/echelle.py
```diff
@@ -2630,43 +2697,102 @@ def line_fit_initial_guess(wavelengths, spectrum, H_a=20, H_b=170, D_a=80, D_b=1
...
     """
+    num_params = 8
+
     # Total flux of H and D in DN, initial guess: get by integrating around the line. Note that the H bounds as defined
     # in a parent function overlap the D, but that's okay for an initial guess.
+    rows = get_n_int(light_fits) if coadded is False else 1
+    guesses = np.zeros((rows, num_params))
+    # Line center of IPH is predicted based on the obs geometry;
+    # length is number of integrations.
+    lambda_IPH_lya_guess = 121.567 + predict_IPH_linecenter(light_fits)
```
planetarymike, line 2870: probably should fit the IPH parameter the same way as D, by fitting the H central wavelength and then subtracting the IPH line center offset from the lab wavelength, but if this works that's fine for now.

maven_iuvs/echelle.py
```diff
@@ -2630,43 +2697,102 @@ def line_fit_initial_guess(wavelengths, spectrum, H_a=20, H_b=170, D_a=80, D_b=1
...
+
+    return guesses
+
+
+def predict_IPH_linecenter(light_fits):
+    """
+    Exact location of upstream direction not completely certain. 
+    Not accounted for: decelreation of IPH across solar system, assumed to be the same.
+    """
+    load_iuvs_spice()
```
planetarymike, line 2933: I think this will clear and load the spice kernels every time this function is called... possibly greatly impacting performance. May be better to check whether the kernels are loaded and return an error if they are not, or put the contents of the function into a try/except block that asks the user to load_iuvs_spice() if a spice error is thrown.

planetarymike (commented):
I did not review changes to the graphics routines.

This looks ready to merge depending on your thoughts about some of my suggestions. I think the main thing that's still missing is ensuring the IPH fit values are reported in the l1c files. We'll need to make some minor changes to the SIS I guess once these are slated for delivery.

maven_iuvs/echelle.py
emcangi: Fixed in 1dc86bfe73b86b1aad026fa19d45487656e827a4

maven_iuvs/echelle.py
emcangi: Fixed in 38c81f7ccf7737d2e3e3e95e56dc52065f78456a

maven_iuvs/echelle.py
emcangi: Oh nooooo, yes

maven_iuvs/echelle.py
emcangi: Ah ok, I misunderstood and did 5 as the total spread, apparently I can't read

emcangi: Fixed in 2b5aa95a4a3260adfed691da8f079d988d183a70

maven_iuvs/echelle.py
planetarymike: I think the new implementation is still redoing all of the work in every loop, but this is probably not an important optimization to make.

maven_iuvs/echelle.py
planetarymike: I think you either need to update the comment in constants to refer to 10 km/s or change the constant back and remove the /2 from the bounds array.

maven_iuvs/echelle.py
planetarymike: let me know if you're planning on making this change or just leaving it as is.

maven_iuvs/echelle.py
planetarymike: Hopefully it's not too much extra work!

emcangi: Fixed with c9545179fb52629ce74c4971ee40e51fe6117a2b

maven_iuvs/echelle.py
emcangi: If we do this, then what do we do with the IPH velocity uncertainty? Currently that's ±5 km/s and that provides the bounds over which to search for the line center.

maven_iuvs/echelle.py
emcangi: Fixed in c9545179fb52629ce74c4971ee40e51fe6117a2b

maven_iuvs/echelle.py
planetarymike: We keep it because it represents the uncertainty in the offset relative to the mars rest wavelength. The issue is whether a "0" fit parameter for the IPH offset within the bounds of +/-5 km/s represents zero offset from the lab frame wavelength of Lyman alpha or from the fitted instrument Mars H Ly a wavelength. I think it should be the latter to minimize potential systematics involved in referencing the IPH to the lab wavelength but fitting the observed location of the Mars H line center.

maven_iuvs/echelle.py
```diff
@@ -2160,11 +2168,19 @@ def fit_H_and_D(pig, wavelengths, spec, light_fits, CLSF, unc=1, IPH_bounds=(Non
...
+                                   args=objfn_args)
 
             if fit_IPH_component:
                 fit_uncert = np.sqrt(np.diag(inv(hessian)))
-            else:
+                # Sometimes we get nans if the epsilon of approx_hess2 is chosen 
+                # automatically. Check and recalculate if need be.
+                if np.isnan(fit_uncert).any():
+                    new_hessian = approx_hess2(bestfit.x, negloglikelihood, 
+                                   epsilon=1e-2, # Seems to be the magic number...
```
planetarymike: Based on the approx_hess2 documentation (available here: https://www.statsmodels.org/dev/generated/statsmodels.tools.numdiff.approx_hess2.html ), the default is to use

`epsilon = EPS**(1/3)*x`

That is, the epsilon used in the hessian computation is a fraction of the fitted parameter value. The default value for EPS is `np.finfo(float).eps`, which on my machine is ~2e-16, giving `epsilon ≈ 6e-6*x` .

Probably we should also scale epsilon by the best fit value, so I recommend trying `epsilon=1e-2*bestfit.x` .

planetarymike:
Pushed some changes to switch the lineshape_model and (neg)loglikelihood to use JAX. This has two advantages:

1. The (neg)loglikelihood function is just-in-time compiled, which reduces function runtime. Fitting with dynesty now takes ~15 s (static sampler) or ~30s (dynamic sampler).
2. The hessian can be computed via automatic differentiation rather than finite differences. Hopefully this results in fewer singular hessian matrices.

I have not tested this code for all modes (in particular, I have not tested the BU_background specification, which needed to be altered for the code to work with JAX). It's possible modes other than the standard mode are broken.

JAX is a bit tricky to work with if conditional evaluation is needed, but it's not too bad, hopefully this doesn't add a ton of overhead when future changes are needed to the lineshape model.

emcangi:
Noting that we tested this together and it looks like the dynesty fitter with the following options works at a ~1s/integration on my machine:
```
convert_l1a_to_l1c(light_fits, dark_fits, light_path, dark_path, l1c_savepath, 
                   fitter="dynesty", approach="static", bound="single", livepts=100)   # <-------------------
```

That's still slower than scipy, I think, but without the problems associated with approx_hess2, so I think it's worth it.

maven_iuvs/echelle.py
emcangi: Yes this seems to have been an oversight in trying to adapt to quicklooks. it has been fixed for real this time in 9756bee44f864647d52fa08bab306de82ab6bdc3

maven_iuvs/echelle.py
emcangi: This is no longer relevant thanks to the jax implementation, thank you!

maven_iuvs/echelle.py
emcangi: Could you possibly add a change for this? I'm still working on error control for problem files.

emcangi: We discussed in person and found that on my machine, this takes ~microseconds, and so determined not to fix this for now. Let's keep an eye on it in case it becomes a problem for us or future users.

maven_iuvs/binning.py
```diff
@@ -143,16 +144,14 @@ def get_pix_range(myfits, which="spatial"):
     pixlo = myfits['Binning'].data[f'{which[:3]}pixlo'][0]
     pixhi = myfits['Binning'].data[f'{which[:3]}pixhi'][0]
 
-    if not (set((pixhi[:-1]+1)-pixlo[1:]) == {0}):
```
planetarymike, line 146: The issue there is that spaepixlo and spaepix high include the indices of bins that are not transmitted, so there needs to be a cross-check against spaebintransmit to see whether or not the bin is actually included in the file. I think at some point in some application of this function I worried about dividing the data by compute number of pixels per bin, and that division won't work if the bin widths are calculated using a diff on this function. Maybe it's OK now.

planetarymike: I took a closer look and the issue is in `get_npix_per_bin()` . For nonlinear binning schemes that function needs to be modified to only report transmitted bins. Probably an easy fix but for now the nonlinear warning can be moved there

planetarymike: It also looks like some of the echelle graphics routines will need to be modified to account for this too if the nonlinear exception is removed...

emcangi: When I tested this, it produced expected output, with the only difference being that the untransmitted bins leave a much larger "gap" in the pixels. E.g., for a file that is linearly binned with width 11 pix/bin in transmitted bins, and 77 pix/bin in two untransmitted bins in the ~middle of the detector, just above the slit:

```python
>>> print(f"file: {lf['primary'].header['FILENAME']}")
file: mvn_iuv_l1a_outdisk-orbit02102-ech_20151029T030345_v13_r01.fits
>>> print(f"Binning scheme: {iuvs.binning.get_binning_scheme(lf)}")
Binning scheme: {'bintbl': 'NON LINEAR 4,4 nonlin_0002', 'nspa': 50, 'nspe': 160}
>>> print("Transmitted/untransmitted bins: ", lf['binning'].data['spabintransmit'])
Transmitted/untransmitted bins:  [[0 0 0 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 0 0 1 1 1 1 1 1
  1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 0 0]]
>>> print(f"Bin widths: {lf['binning'].data['spabinwidth']}")
Bin widths: [[100 100  99  11  11  11  11  11  11  11  11  11  11  11  11  11  11  11
   11  11  11  11  11  11  11  11  11  11  77  77  11  11  11  11  11  11
   11  11  11  11  11  11  11  11  11  11  11  11  11  11  11  11  11  11
   11  17   4]]
>>> print(f"Lowest pixel number in bin: {lf['binning'].data['spapixlo']}")
Lowest pixel number in bin: [[299 310 321 332 343 354 365 376 387 398 409 420 431 442 453 464 475 486
  497 508 519 530 541 552 563 728 739 750 761 772 783 794 805 816 827 838
  849 860 871 882 893 904 915 926 937 948 959 970 981 992]]
>>> print(f"Highest pixel number in bin: {lf['binning'].data['spapixhi']}")
Highest pixel number in bin: [[ 309  320  331  342  353  364  375  386  397  408  419  430  441  452
   463  474  485  496  507  518  529  540  551  562  573  738  749  760
   771  782  793  804  815  826  837  848  859  870  881  892  903  914
   925  936  947  958  969  980  991 1002]]
>>> print(f"results of get_pix_range(): {iuvs.binning.get_pix_range(lf, which='spatial')}")
results of get_pix_range(): [ 299  310  321  332  343  354  365  376  387  398  409  420  431  442
  453  464  475  486  497  508  519  530  541  552  563  574  739  750
  761  772  783  794  805  816  827  838  849  860  871  882  893  904
  915  926  937  948  959  970  981  992 1003]
```

From what I can tell get_pix_range() is only used in: 
- echelle.py
- echelle_graphics.py
- geometry.py (one call, and it's not even imported, so I think it must not have ever been used?)
- line_fit_plot.py

From what I can tell in my testing, it's working in all of those except geometry.py, but I haven't tested it for use with the standard resolution. Is there another case you're thinking of where this could break?

Also, these commits weren't supposed to be on this branch since they're not strictly related to IPH. I can look more into moving them to another branch later, at a brief glance it looked like it was going to be a bit of a headache.

planetarymike: `get_pix_range()` is called by `get_npix_per_bin`, which is called by `detector_image`. In that routine the following happens (this is similar to some of your echelle routines I think):

```python
def detector_image(myfits, integration=0, **kwargs):
    data = myfits['detector_dark_subtracted'].data[integration]

    spapixrange = get_pix_range(myfits, which="spatial")
    spepixrange = get_pix_range(myfits, which="spectral")
    npixperbin = get_npix_per_bin(myfits)
            
    data = data / npixperbin

    fig = plot_detector(data, spapixrange, spepixrange, **kwargs)
```

If get_pix_range returns extra entries for non-transmitted bins, there will be a dimensional mismatch between the `data` element and `npixperbin`. Additionally, `plot_detector()` will have issues because the `pcolormesh` call will have a dimensional mismatch between the data and the pixel boundaries given. 

If we want to retain the new behavior of `get_pix_range()`, that's OK, but the downstream routines would either need to be fixed to operate correctly with nonlinear binning or raise an error similar to the old error raised by `get_pix_range()` in the event a file with nonlinear binning is encountered.

planetarymike: Maybe I'm still missing something--- if so we should chat in person!

maven_iuvs/echelle.py
emcangi: After discussion, I implemented this in 6795164877bbaa7fd02e2528f26cfc204b03e771, but found that it adds a huge amount of overhead to the fitting for some reason - 5-10 seconds per call, which is not sustainable. Thus I've reverted it in 4fa5e9b4d53c3d383802df8f5c6abeb51708496c and for now we'll just accept that there will be a slight discrepancy that should be accounted for within the uncertainties anyway.

maven_iuvs/binning.py
emcangi: Its probably hard to discuss in thread but:

> If get_pix_range returns extra entries for non-transmitted bins, 

I was saying this doesn't happen as far as I can tell, at least when run on echelle files. In my snippet above, in the last line, it jumps from 574 pix to 739 whereas the rest are diff'ed by 11 pix only. I was able to make quicklooks with it (as I showed you in person, and I have pasted it below again to link it to this discussion).


image


I checked every entry in the index file where the binning is nonlinear, and they all have the same scheme, so they all should work (not gonna test it because there's 2400+ files and it will take forever). 

I can still run `detector_image()` / `plot_detector()` successfully on a linearly-binned l1b FUV file. I'm happy to test on a nonlinearly binned l1b FUV file if there were any (thinking maybe not? I went back to the range where echelle was being nonlinearly binned, and those FUV files were linearly binned), but I'm less familiar with it and don't have it downloaded.

I suppose it's possible in the future we could change the binning again, so we should build in a catch? Would that satisfy this thread?

emcangi: Okay, I've added an assertion in the detector image routines here: d6d354e

Let me know if that looks good enough to complete the review!

planetarymike: I just pushed some changes to fix the remaining issues.

Nonlinearly binned files could be handled pretty well by the previous code, but the issue was that one transmitted pixel would be assigned an incorrect dimension that extended over any non-transmitted bins to the lower edge of the next transmitted bin. Now this does not occur--- binning functions have been modified to return values that are correct regardless of linear / nonlinear binning, and plotting functions have been modified to pad data arrays with NaN values for nontransmitted bin ranges.

emcangi (commented):
LGTM

maven_iuvs/binning.py
emcangi: Great!

planetarymike (approved):
I think this is ready to go!

emcangi:
Just confirming - I went back and checked the file where the H line was being severely underfit with the IPH updates. That file is now fine, so I think the issue was related to the lack of geometry - that or other changes we made in the process fixed the problem.

(file was: mvn_iuv_l1a_inlimb-orbit24112-ech_20250801T032926_v13_XYY.fits.gz, where x= r|s and y=\d)

PR #82, opened 2025-08-06
Migrate pkg resources

### LINKED ISSUE #81: Replace pkg_resources, which is deprecated
```
[.../maven_iuvs/maven_iuvs/geometry.py:7](.../maven_iuvs/maven_iuvs/geometry.py#line=6): UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
```
Used in just one place (line 188 of geometry.py). Exact role needs to be determined and then it should be replaced with something that's not deprecated.

planetarymike:
## Pull Request Checklist
 - [N/A] New routines don't duplicate existing functionality

Fixes #81

### DIFF
--- a/maven_iuvs/__init__.py
+++ b/maven_iuvs/__init__.py
@@ -19,7 +19,7 @@ from . import file_classes
 from . import binning
 from . import echelle
 from . import reprocessing
-from . import compare_pipelines
+# from . import compare_pipelines
 from . import pds
 
 try:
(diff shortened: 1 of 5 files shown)


emcangi (approved):
LGTM but I'm seeing some conflicts, will take a look

Issue #50, opened 2023-05-09
Review module for obsolete / unused code


planetarymike:
There is some legacy code from Kyle and Zac that Mike and Eryn aren't using, it would be good to prune this for clarity.

PR #83, opened 2025-08-12
Upgrades to dark handling, parallel processing, and adding model of correlation between wavelength bins

emcangi:
## Pull Request Checklist
 
This code mainly includes:
- Update function that creates a .csv file that links lights with darks
- Add a new function that searches a light/dark key .csv file and determines which entries may have had a revision update since the file was made, and returns a new dataframe containing all the entries including updated ones. (must be written to .csv in a higher level call that's not included here--if it makes sense to include it here, I will make a new commit). 
- Improved readability and logic behind determining when dark frames are usable or not, and what to do with the related light frames

I am the original writer of all the routines but would especially like feedback from @planetarymike on 0ce26822926357d40eadd18072617b6873e509fa.

Other changes included here are not large or super important, but helpful for me to merge before completing a new mission reprocess.


maven_iuvs/echelle.py
```diff
@@ -657,21 +657,43 @@ def subtract_darks(light_fits, dark_fits):
     first_dark = darks[0, :, :]
     second_dark = darks[1, :, :]
 
-    # check darks are valid
-    if np.isnan(first_dark).any() & np.isnan(second_dark).any():
-        raise Exception(f"Missing critical observation data: no valid darks")
-
```
emcangi, line 599: Please note this logic is still present, but improved to catch more possible reasons they may be bad, see code at right, line 694

.github/workflows/github-actions-demo.yml
```diff
@@ -0,0 +1,18 @@
+name: GitHub Actions Demo
```
planetarymike: Is this file really supposed to be a part of this PR? Recommend deletion.

maven_iuvs/search.py
```diff
@@ -337,6 +339,8 @@ def get_latest_files(files):
     Prefers highest version number, then production files to stage files,
     and finally highest revision number.
 
+    NOTE: This only works if multiple versions of the files are in the list.
```
planetarymike: I think if multiple versions are not present, this should just return the original list, right? So it 'works', there's just nothing to do.

maven_iuvs/search.py
```diff
@@ -321,6 +320,9 @@ def get_file_version(orbit_number, data_directory,
                            segment=segment,
                            channel=channel,
                            data_directory=data_directory)
+        # TODO: The above call returns a list of strings, but the
```
planetarymike, line 400: Actually, this whole function is gross because it doesn't operate on a file at all but goes to disk to look at what's there. I don't think this function is called by anything else in the repo, recommend deletion.

maven_iuvs/echelle.py
```diff
@@ -334,11 +334,163 @@ def downselect_data(index, light_dark=None, orbit=None, date=None, segment=None,
...
+    problems : pandas DataFrame
+               Files that caused a problem
+    """
+    MASTER_KEY = pd.read_csv(keyfile)
+
+    changes = {"row": [], "oldname": [], "newname": []}
+    problems = {"row": [], "oldname": [], "newname": []}
+
+    # Precompile regex
+    versioned_pat = re.compile(r"^(?P<base>mvn.+?)_(r|s)(?P<rev>\d+)(?P<ext>\.fits\.gz)$")
```
planetarymike: you're actually gonna make everyone who maintains this repo have to learn regular expressions, aren't you??? :sob: :cries due to laziness:

maven_iuvs/echelle.py
```diff
@@ -334,11 +334,163 @@ def downselect_data(index, light_dark=None, orbit=None, date=None, segment=None,
...
+        rev = int(lmatch.group("rev"))
+        ext = lmatch.group("ext")
+
+        # Find all matching versions in the folder
+        lmatches = []
+        for f in light_folder_cache[lfolder]:
+            m = versioned_pat.match(f)
+            if m:
+                f_base = m.group("base")
+                f_rev = int(m.group("rev"))
```
planetarymike: how do you handle stage vs. production files?

maven_iuvs/echelle.py
```diff
@@ -334,11 +334,163 @@ def downselect_data(index, light_dark=None, orbit=None, date=None, segment=None,
...
+        if lmatches:
+            newest_file = max(lmatches, key=lambda x: x[1])[0]
+            if original_lightname != newest_file:
+                new_lightnames.loc[i] = newest_file
+                changes["row"].append(i)
+                changes["oldname"].append(original_lightname)
+                changes["newname"].append(newest_file)
+
+        # DARKS ------------------------------
+        if dfolder not in dark_folder_cache:
```
planetarymike: dark handling code copy pasted from light handling code above is begging for encapsulation, but if it works it works.

maven_iuvs/echelle.py
```diff
@@ -545,13 +708,15 @@ def coadd_lights(data, n_good):
            detector image data, size (frames, spatial, wavelength)
     n_good : int
              number of valid frames of data used, calculated by subtract_darks.
-
+    exclude_frames : list
```
planetarymike: does not seem to be a parameter in definition above?

planetarymike: also missing in function below... is the docstring change wrong?

maven_iuvs/echelle.py
planetarymike: line numbers don't match but I see code below under the comment `CHECK VALIDITY OF DARKS`

maven_iuvs/echelle.py
```diff
@@ -588,26 +753,50 @@ def subtract_darks(light_fits, dark_fits):
...
+    second_dark_good = True
+
+    # # Check goodness of first dark 
+    if (np.isnan(first_dark).any()) or (np.nanmedian(first_dark) > 1000):  # First dark is bad 
+        first_dark_good = False
+    
+    # Check that second dark exists - early in mission, we didn't always take 2 darks.
+    # In those cases we should use the first dark.
+    second_dark_exists = True 
+    if np.isnan(second_dark).all():
```
planetarymike: add a comment: `# an all nan second_dark is guaranteed by get_dark_frames() above if only one dark frame exists`

maven_iuvs/echelle.py
```diff
@@ -636,22 +825,23 @@ def subtract_darks(light_fits, dark_fits):
...
+    # Mark frames as bad if they are associated with the first dark, either by
+    # BEING the first dark, or being the first light, or being the 1-nth light
+    # if the second dark doesn't exist.
+    if not first_dark_good:
         nan_dark_inds.append(0)
         light_frames_with_nan_dark.append(0)
-    else:  # First dark is good 
-        if not second_dark_exists: 
+
+        # A bad 1st dark can't be used for light frames 1-n_int in early files
```
planetarymike: I'm not sure we determined that the first dark issue was so bad that we should throw out all the light frames for which no appropriate dark exists. My read of the matching code in the IUVS IDL pipeline suggests that the first dark is used for all frames when no second frame exists.

Dark determination is in the following nested routines:
- `code/superdriver/driver_code/maven_iuvs_process_l1a_to_l1b.pro`
- `code/level1b/iuvs_to_fits_1b.pro`
- `code/level1b/stack_1b_calibrate.pro`
- `code/level1b/iuvs_estimate_dark.pro`

Probably not worth changing to use the IDL procedure now, but maybe worth opening an issue. How many integrations are loss with this treatment?

maven_iuvs/echelle.py
```diff
@@ -2005,7 +2196,7 @@ def writeout_l1c(light_l1a_path, dark_l1a_path, l1c_savepath, light_fits, fit_pa
                                 stdout=outputfile, stderr=errorfile, 
                                 text=True, bufsize=1)
         proc.stdin.write(".com write_l1c_file_from_python.pro\n")
-        time.sleep(3) # Be sure it's compiled 
+        time.sleep(5) # Be sure it's compiled
```
planetarymike: should probably read the output from the terminal to ensure the process has ended, but this can be a todo

planetarymike (commented):
Just a few small comments, hopefully super easy to address.

maven_iuvs/echelle.py
emcangi: https://xkcd.com/208/

.github/workflows/github-actions-demo.yml
emcangi: No, I was just being lazy

maven_iuvs/search.py
emcangi: Must have been used by Kyle? But now they have a whole other repository. Anyway, I actually think we should leave it for now. This falls under the issue #50 and I think it should have a separate PR (i.e. we can work on it later). that ok with you?

planetarymike: OK with me to leave it for now

maven_iuvs/echelle.py
planetarymike: Does it prefer production to stage? Would it prefer _r01 to _s03, for example?

maven_iuvs/echelle.py
emcangi: added in 84cb3f619af0f201db32140ec823d57714f2519c

maven_iuvs/echelle.py
emcangi: I think very few files have that problem. I agree with you though, I think we should just use the 1st dark. I didn't code that up yet out of caution. It can be an issue.

maven_iuvs/echelle.py
emcangi: This function has been fully rebuilt to address this in 43aca59752dfa8dcfe3b41ed84d20b0d3feb7087

maven_iuvs/echelle.py
emcangi: Well, it was initially treating the copy on disk as the the copy to prefer. But I think that could backfire in rare circumstances. So I rebuilt the whole thing to make it more efficient, compare r vs. s and numbers of revision, and not duplicate code. 43aca59752dfa8dcfe3b41ed84d20b0d3feb7087

maven_iuvs/echelle.py
emcangi: I added some cheat codes c81201442917ba32f66d598d3376813f2fc7d6b6

maven_iuvs/search.py
emcangi: Clarified what I meant 81300aef8b8b0da8e914f3aa17b9ab02d1a8b2ab here

.github/workflows/github-actions-demo.yml
emcangi: 425c02f3191cc8329a1472153c6b7504d2a7e92b

maven_iuvs/echelle.py
emcangi: Hmmm yeah I think this is an artifact. I think it's not needed because I do nansum to ignore any bad values, and n_good should be calculated correctly. I've been testing various quicklooks and they all looked okay. I'll double check tomorrow but for now I've removed this in 2deddcd9215b5960a570053b65ca16ec583551a1

maven_iuvs/echelle.py
emcangi: It has to be read from stdout or stderr or something and it was very hard to get this all working at all, so definitely a todo item

maven_iuvs/echelle.py
emcangi: It now compares r and s, preferring r, and then just picks the biggest revision number available. You may want to have a quick look at the update. I only check to see if the disk file is newer in one of those ways - if the file listed in the light/dark key csv is newer than whatever the user has on disk, nothing is done. (but I don't know how that would be different anyway. It would be weird to ask that routine to go download new files!)

maven_iuvs/echelle.py
emcangi: (all ordinals mentioned in this comment assume zero-indexed)

Wait a second - what the comment line you put this comment on is saying is: if there is no 1st dark frame, AND the 0th dark frame is bad/unusable, then there are no valid dark frames. 

Further down in the code, the 0th dark frame---if it's valid and usable---IS used for the 1st through nth light frames if there was only 1 dark frame taken total. 

I think the only remaining issue is in observations with 2 dark frames. In that scenario, if dark 0 is good and dark 1 is bad, lights 1 through n are thrown out. In that case, should we go ahead and use the 0th dark?

maven_iuvs/search.py
planetarymike: Makes sense, good to clarify the docstrings here.

maven_iuvs/echelle.py
planetarymike: I still feel like this should be able to rely on `get_latest_files` somehow, but it's OK to leave this functionality as separate for now.

maven_iuvs/echelle.py
planetarymike: where is the 0th dark used? If all of the integers and comments are zero-based, should we change the comments and variable names to reflect this to avoid confusion, or would that only introduce more confusion?

planetarymike: OK, I see it on line 879 below. All good.

planetarymike:
lmk when you're done with this PR--- it looks like you're still committing some changes.

emcangi:
@planetarymike Yes, realized I hadn't added the fitted values to the data products, so had to do that. I also decided to add the ability to watch the IDL output to check for completion now... anticipating that if I didn't, I would get tons more hard-to-analyze errors across the mission and have to go back and fix them all, creating more work. 

FYI, the performance has really tanked--up to 3 minutes to generate a single file sometimes. It may be due to the changes made to allow for checking that the script is compiled, etc. There is also the issue of now that we are saving all the model fits to a new HDU, I have it writing out like 5 or 6 .csv files from python, then loading them in IDL. If I had to guess it's probably the first thing slowing it all down - I'll have another go at it later, but for now I'll just let it run slowly. So far it's taken about an hour to do orbits 100-400. Usually it takes minutes.

emcangi:
Had a look at the performance and it looks like the fitting dominates things, and dynesty is still an order of magnitude slower than scipy:

```
# Fit an observation on orbit 724 with all broken or nan frames - Scipy
2025-08-19 11:04:24 Start (t0)
2025-08-19 11:04:24 Dark subtraction: 0.01 seconds (t1) 
2025-08-19 11:04:25 Artifact removal: 0.98 seconds (t2) 
2025-08-19 11:04:27 Finished flattening and fitting: 2.13 seconds (t3)  # <-----------------------
2025-08-19 11:04:27 Opening IDL
2025-08-19 11:04:27 Finished opening IDL and compiling script: 0.16 sec (tb-ta)
2025-08-19 11:04:28 Finished with IDL writeout: 0.45 sec (td-tc)
2025-08-19 11:04:28 Finished writeout with IDL: 0.64 seconds (t4)
```

```
# Fit an observation on orbit 724 with all broken or nan frames - Dynesty
2025-08-19 11:04:28 Start (t0)
2025-08-19 11:04:28 Dark subtraction: 0.01 seconds (t1) 
2025-08-19 11:04:29 Artifact removal: 1.0 seconds (t2) 
2025-08-19 11:05:05 Finished flattening and fitting: 35.83 seconds (t3) # <-----------------------
2025-08-19 11:05:05 Opening IDL
2025-08-19 11:05:05 Finished opening IDL and compiling script: 0.16 sec (tb-ta)
2025-08-19 11:05:05 Finished with IDL writeout: 0.45 sec (td-tc)
2025-08-19 11:05:05 Finished writeout with IDL: 0.64 seconds (t4)
```

```
# Fit an observation on orbit 21494 - Scipy
2025-08-19 10:55:05 Start (t0)
2025-08-19 10:55:05 Dark subtraction: 0.01 seconds (t1) 
2025-08-19 10:55:06 Artifact removal: 1.01 seconds (t2) 
2025-08-19 10:55:08 Finished flattening and fitting: 2.17 seconds (t3)  # <-----------------------
2025-08-19 10:55:08 Opening IDL
2025-08-19 10:55:42 Finished opening IDL and compiling script: 0.17 sec (tb-ta)
2025-08-19 10:55:43 Finished with IDL writeout: 0.47 sec (td-tc)
2025-08-19 10:55:43 Finished writeout with IDL: 0.67 seconds (t4)
```

```
# Fit an observation on orbit 21494 - Dynesty
2025-08-19 10:56:27 Start (t0)
2025-08-19 10:56:27 Dark subtraction: 0.01 seconds (t1) 
2025-08-19 10:56:28 Artifact removal: 0.99 seconds (t2) 
2025-08-19 10:57:04 Finished flattening and fitting: 36.56 seconds (t3)  # <-----------------------
2025-08-19 10:57:04 Opening IDL
2025-08-19 10:57:05 Finished opening IDL and compiling script: 0.16 sec (tb-ta)
2025-08-19 10:57:05 Finished with IDL writeout: 0.48 sec (td-tc)
2025-08-19 10:57:05 Finished writeout with IDL: 0.66 seconds (t4)
```

I'll try 'lazy parallelizing' for now in my bash script that calls Python, since I also still get NaNs in my uncertainties when fitting with Python (tried many combinations of xtol and ftol). With that in mind I think you're good to review this.

emcangi:
Much has been committed but I think this is in a good state for updated review. Unfortunately, fixing issues with darks just led to more and more rabbit holes, and as the pipeline evolved, treatment of darks had to evolve as well in complex ways. 

Here's what's been updated:

1. Dark identification routines and pairing routines have been fixed multiple times.
2. Use of dark frames (e.g. when one or more is bad, when a light frame is bad, etc) has been improved
3. Significant updates to routines that update and keeps updated the light/dark pair files for the whole mission
4. Model fitting results are now written to L1c files, both in physical units (kR/nm) and DN. 
5. The whole pipeline now supports multiprocessing, instigated by RUN_ECHELLE_v15_PIPELINE.py.
6. Transfer of information from Python to csvs that IDL picks up has been finalized.
7. Wavelength grid for fitting is now constructed based on dispersion relation, rather than using what's listed in the l1a files, which per Justin is probably not reliable, and are written out to the l1c files.
8. A function is now available that identifies when the IUVS look direction is well out of the orbital plane.
9. Weekly reports now report data present on even and odd orbits
10. Dynesty processing speed improved to be about equal to that of Scipy processing
11. Explicit functions for chisquared have been added

Overall commits are fairly short so viewing each commit may make the most sense. There is one commit that reorganized code in the middle, so that may make it hard to view everything at once.

maven_iuvs/echelle.py
```diff
@@ -3079,6 +3092,55 @@ def make_fit_param_dict(thelist, is_fitparams=True):
...
+    padded_kernel = np.pad(kernel, (0, n_wave_bins-len(kernel)))
+    # Mirror the kernel at the back end of this vector so that the first 
+    # column in the array will have a 1 in the top left, allowing us to put 
+    # 1's along the diagonal.
+    padded_kernel[-(len(kernel)-1):] = np.flip(kernel)[:-1]
+    kernel_array = np.zeros((n_wave_bins, n_wave_bins))
+    # Fill each column of array with the full kernel vector, shifted by i 
+    # places toward the "right" (or end). 
+    for i in range(len(padded_kernel)):
+        kernel_array[:, i] = np.roll(padded_kernel, i)
```
planetarymike: This, in combination with the padded_kernel command on line 3136, will introduce correlation between one side of the spectrum and the other due to the presence of non-zero elements outside the band diagonal. The non-zero elements in the upper right and lower left corners of the matrix need to be set to zero after constructing the correlation matrix with this method.

emcangi: Fixed with https://github.com/lasp/maven_iuvs/pull/83/commits/f6102cec92f445ad56470fd0c5eb7ee18813d7ec

maven_iuvs/instrument.py
```diff
@@ -321,9 +330,9 @@ def ran_DN_uncertainty(light_fits, dark_subtracted_and_cleaned_data):
 
     # This is the correct shape, not sure if it's reasonable values though:
     ran_DN_sq = dark_subtracted_and_cleaned_data * fit_function + sigma_background**2
-    # if np.any(ran_DN_sq < 0.):
-    #     raise ValueError("negative values in ran_DN_sq")
-    ran_DN = np.sqrt(ran_DN_sq)  # TODO: check if it's okay to do this on the cleaned data. Probably not
-    ran_DN[np.where(np.isnan(ran_DN))] = 0  # TODO: this is not acceptable lol
+
+    ran_DN = np.sqrt(ran_DN_sq)
+    # Nans can occur in the dark_subtracted_and_cleaned_data term, but that's okay.
```
planetarymike: I’m not sure how a NaN results from the calculations above? Doesn’t seem like any of the functions above can produce a negative value

maven_iuvs/miscellaneous.py
```diff
@@ -69,12 +74,22 @@ def mirror_dn_to_deg(dn, inverse=False):
...
+def find_nearest(array, value, price_is_right=False):
     """
     Find the closest entry in array to value. 
+
+    Parameters
+    array : array
+    value : float
+    price_is_right : bool
+                     If true, finds the nearest entry to v in array that isn't 
+                     larger than value.
```
planetarymike: isn’t this argsort?

maven_iuvs/search.py
```diff
@@ -1,18 +1,94 @@
...
 #  NOTE: depends on maven_iuvs.download must be encapsulated to avoid
 #  circular import
 from maven_iuvs.geometry import beta_flip
+from maven_iuvs.miscellaneous import uniqueID_RE, orbit_folder, iuvs_orbno_from_fname
+
+
+def build_cache(orbfold=None, cache={}, exclude_xml=True, exclude_dark=True, exclude_txt=True,
+                basefolder="/media/emc/ExtremePro/IUVS/IUVS_Data/l1c_ech_data/FMR_v15/DynestyAWS/"):
+    """
+    Given an orit folder orbfold, will fill the cache with all the files there.
```
planetarymike: typo

maven_iuvs/statistics.py
```diff
@@ -1,5 +1,35 @@
...
+        reduced : bool
+                  False > returns normal chi squared. True > returns reduced
+
+    Returns
+    ----------
+    chisq : float
+            chi squared value, normal or reduced.
+    """
+
+    chisq = -2 * (LL + (N/2)*np.log(2*math.pi) + 0.5 * logdetcov)
```
planetarymike, line 29: does chi-squared extend to multivariate normals will non-zero correlation? Or is it necessarily for independent variates?

maven_iuvs/echelle.py
```diff
@@ -892,104 +1201,434 @@ def make_dark_index(ech_l1a_idx):
...
+        # Early mission, only one dark frame was taken.
+        darks[0, :, :] = dark_fits['Primary'].data[0]
+        darks[1, :, :] = np.nan # Set the second frame to nans if there was only one dark frame taken
+    elif n_ints_dark == 2:
+        # Noise pattern of the first and every other frame is different. Eventually, we realized this
+        # and started taking two darks
+        darks[0, :, :] = dark_fits['Primary'].data[0]
+        darks[1, :, :] = dark_fits['Primary'].data[1]
+    elif n_ints_dark > 2:
+        # If there's more than 2, we can just take the element-wise mean of frames 2:end. Ignore nans.
```
planetarymike, line 1350: Should we avoid NaNs? That is, do NaNs possibly indicate a problem with the dark that means we should throw it out?

maven_iuvs/echelle.py
```diff
@@ -2313,114 +2872,443 @@ def prior_transform(u, param_bounds):
...
+        kernel_array[:, i] = np.roll(padded_kernel, i)
+
+        # Mirror values above diagonal
+        if i <= len(kernel)-1:
+            mirrored_partial_kernel = np.flip(kernel[1:1+i])
+            kernel_array[:i, i] = mirrored_partial_kernel
+        if i >= len(kernel):
+            # After a certain point we have to add more zeros
+            mirrored_kernel_padded = np.flip(np.pad(kernel[1:], (0, i-len(kernel[1:]))))
+            kernel_array[:i, i] = mirrored_kernel_padded
```
planetarymike: This method of construction seems much more confusing to understand than the for loop I used in the original notebook, are there really performance considerations that require us to use this more opaque method of construction?

planetarymike (approved):
A few comments, but I didn’t do a detailed review of most of the changes in echelle.py, particularly the dark identification code. If you want me to look at that in detail, let me know!

maven_iuvs/miscellaneous.py
emcangi: Unsure about how argsort applies, but I realized there's two extraneous steps here. Reduced the unneeded code and fixed the string which referred to nonexistent variable v. Commit: 58af87d7176e7e3ccae99f7757995b9a01b7bb5f

maven_iuvs/instrument.py
emcangi: Clarified a lot of docstrings and comments in 1d2c76adc180d889b02514470acc7bb4ca63c7e1 to support this. tl;dr here:

Within this specific function, I believe NaNs can appear if for some reason the ran_DN calculation produces a negative value for ran_DN_sq. The equation is

```
ran_DN_sq = dark_subtracted_and_cleaned_data * fit_function + sigma_background**2
```
fit_function and sigma_background should always be positive values so I believe this can only be negative if some bin within dark_subtracted_and_cleaned_data is negative and with an abs value >  fit_function + sigma_background**2 at the same bin. I don't know how often or if this happens, but it may. 

Outside this function, the pipeline is set up so that any light frames that are 'bad' (not usable in some way, i.e. oversaturated, commanding errors, some bins with NaN DN, no usable associated dark frame) are set to NaN after dark subtraction within the `subtract_darks` function*. This was the easiest way to fit while ignoring garbage frames -- specifically it let me write the code to do all the various operations on the entire data cube without having to do 1) tedious for loops or 2) constantly changing array slicing to ignore the garbage frames.

*Note that it currently doesn't do this in the quicklook code, but I think it probably should do it there also. I haven't reexamined the QL code in a while.

maven_iuvs/search.py
emcangi: Haha. 543f552248076afc81e9041c363ebe3b93e0e51e

maven_iuvs/statistics.py
emcangi: After thinking about this, I think yes, chi squared with that name is defined to only apply to independent multivariate normal distributions. What I've done here is solve for some quantity that relates the data to the model in the likelihood that we use, and called it chi squared (since it used to be that we were assuming independent gaussian variates). So the name may not be appropriate anymore, but I don't at present know a better way to quantify goodness of fit. 

I think your comment would be addressed by just renaming this 'goodness_of_fit' or similar. Was that your intention?

maven_iuvs/echelle.py
emcangi: Probably a good idea to be consistent, fixed here: b2ec21d1dae6191500a1df2c1f09c97aa55d3d0e

maven_iuvs/statistics.py
emcangi: Now covered here e28bcc37fa008cd2acac088be93266ca493b54d1 after we uncovered that chi squared does apply to non independent variates

maven_iuvs/echelle.py
emcangi: No, I think this was a poor fix for the old version. I tested it, your for loop, and the new one in ad4f70a8ee8d32a2e4b6ec288bf312580ac443dc and the new one is the fastest and hopefully simplest to read:

For loop: 2.8 ms ± 34 μs
Old version in 8708a9dcf765a4f0e7ea0533c9bfa49ba1fc640e: 3.92 ms ± 11.8 μs
New version in ad4f70a8ee8d32a2e4b6ec288bf312580ac443dc: 1.47 ms ± 29 μs 

I also made sure the result was the same across all methods. Let me know if this is acceptable! Either way a difference of 1-2 ms on each file is < 1 sec differences across the full mission, so no, no performance requirements here.

